PR #36, opened 2014-12-17
added functions quote and surround to strings

k7sleeper:
Unfortunately I cannot set-up the development environment under Windows. Thus, I cannot provide tests :-(


k7sleeper:
Thanks!

PR #40, opened 2014-12-22
strings.__all__ as tuple instead of list

k7sleeper:
It's more leightweight and not modifiable from outside.


dgilland:
I went ahead and made this change in the other files via 7a7d5d053025abed3e0e5341ec3e19653616490e.

k7sleeper:
Thanks.

PR #57, opened 2015-01-23
Continued work on v3 features (II)

k7sleeper:


k7sleeper:
As far as I see, feature 'None-safety of string functions' should be complete.

dgilland:
Thanks!

Issue #64, opened 2016-09-21
default value for max_, min_


AnsonT:
It'll be very useful to be able to pass in default value for empty collections  (like Python3)
coll = []
max_(coll, callback=lambda x: x*x, default=0)

currently, it'll error with max() arg

PR #65, opened 2016-09-21
add optional 'default' parameter for max_/min_ function 

AnsonT:
https://github.com/dgilland/pydash/issues/64

>>> max_([], default=-1)
-1

### DIFF
--- a/pydash/collections.py
+++ b/pydash/collections.py
@@ -62,6 +62,8 @@ __all__ = (
     'where',
 )
 
+__default_param__ = object()
+
 
 def at(collection, *indexes):  # pylint: disable=invalid-name
     """Creates a list of elements from the specified indexes, or keys, of the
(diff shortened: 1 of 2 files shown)


pydash/collections.py
```diff
@@ -556,12 +558,13 @@ def mapiter(collection, callback=None):
...
 
 
-def max_(collection, callback=None):
+def max_(collection, callback=None, default=__default_param__):
     """Retrieves the maximum value of a `collection`.
 
     Args:
         collection (list|dict): Collection to iterate over.
         callback (mixed, optional): Callback applied per iteration.
+        default: default value when collection is empty
```
dgilland: Minor: `default (mixed, optional): ...`

AnsonT:
I added an iterator that returns a default value if empty:

```python
max(_iterator_with_default(collection, default), key=pyd.iteratee(callback))
```

The iterator code is more complex, but doesn't use global, and exceptions all handled by the min/max implementation

```python
class _iterator_with_default(object):
    def __init__(self, collection, default):
        self.iter = iter(collection)
        self.default = default

    def __iter__(self):
        return self

    def next_default(self):
        ret = self.default
        self.default = NoValue
        return ret

    def __next__(self):
        ret = next(self.iter, self.next_default())
        if ret is NoValue:
            raise StopIteration
        return ret

   next = __next__
```

PR #67, opened 2016-10-15
Added strings.lower_case() method

bharadwajyarlagadda:


bharadwajyarlagadda:
@dgilland

pydash/strings.py
```diff
@@ -736,6 +736,56 @@ def lines(text):
...
+def lower_case(text):
+    """Converts string to lower case as space separated words. Before
+    converting the given text to lower case this method:
+
+      - replaces special character with white spaces,
+      - adds a white space before the capital letter,
+      - replaces multiple whitespaces with a single whitespace,
+      - pertains spaces between words in the given text,
+      - pertains the digits given in the text.
+
```
dgilland: I don't think we need to go into this much detail in describing what the function does. The Lodash description is simply `Converts string, as space separated words, to lower case.` which I think would be fine here. Since we have the `words()` function, you could link to it to give more context for what a "word" is.

pydash/strings.py
```diff
@@ -736,6 +736,56 @@ def lines(text):
...
+    Example:
+
+        >>> lower_case('fooBar')
+        'foo bar'
+        >>> lower_case('--foo-Bar--')
+        'foo bar'
+        >>> lower_case('/?*Foo10/;"B*Ar')
+        'foo10 b ar'
+    """
+    # TODO: Add release version.
```
dgilland: I would add this in the docstring as:

```python
"""
.. versionadded:: TODO
"""
```

tests/test_strings.py
```diff
@@ -350,6 +350,20 @@ def test_lines(case, expected):
...
 @parametrize('case,expected', [
+    ('fooBar', 'foo bar'),
+    ('--foo-Bar--', 'foo bar'),
+    ('*Foo*B_a*R', 'foo b a r'),
+    ('*Foo10_B*Ar', 'foo10 b ar'),
+    ('/?*Foo10/;"B*Ar', 'foo10 b ar'),
+    ('/?F@O__o10?>.B?>";Ar', 'f o o10 b ar'),
+    ('FoO Bar', 'fo o bar'),
+    ('F\n\soO Bar', 'f so o bar')
+])
```
dgilland: I would ran all these examples through Lodash because a few of them don't match.

CHANGELOG.rst
```diff
@@ -4,6 +4,9 @@ Changelog
 =========
 
 
+- Add ``lower_case``.
```
dgilland: Go ahead and add `Thanks bharadwajyarlagadda_!` after the changelog entry (see other examples in the file). And at the bottom of the page add:

```rst
.. _bharadwajyarlagadda: https://github.com/bharadwajyarlagadda
```

Gotta give credit where credit is due :+1:

pydash/strings.py
```diff
@@ -736,6 +736,56 @@ def lines(text):
...
+            characters.append(char)
+
+    text = ''.join(characters)
+    text = text.strip()
+
+    # Replace multiple whitespaces between the words in the text with a single
+    # whitespace.
+    text = re.sub("\s\s+", " ", text)
+
+    return text.lower()
```
bharadwajyarlagadda: That wouldn't work. There are several cases to deal with. 
1) We need to add whitespaces after and before a digit 
2) We need to replace all the special characters in the text with a whitespaces
3) We need to add space before the capital letter in the text (before lowering the text.

`words()` can't help with all these cases. And also, I would avoid giving complex regex's which would give a lot of false positives or false negatives.

dgilland: Which edge case doesn't `words()` handle?

```python
>>> from pydash import words
>>> def lower_case(text):
...     return ' '.join(words(text)).lower()
... 
>>> tests = (('fooBar', 'foo bar'),
... ('--foo-Bar--', 'foo bar'),
... ('*Foo*B_a*R', 'foo b a r'),
... ('*Foo10_B*Ar', 'foo 10 b ar'),
... ('/?*Foo10/;"B*Ar', 'foo 10 b ar'),
... ('/?F@O__o10456?>.B?>";Ar', 'f o o 10456 b ar'),
... ('FoO Bar', 'fo o bar'),
... ('F\n\soO Bar', 'f so o bar'),
... ('Foo1054665Bar', 'foo 1054665 bar'))
>>> for case, expected in tests:
...    lower_case(case) == expected
... 
True
True
True
True
True
True
True
True
True
```

The purpose of `words()` is to do all the things you mention. It may be a complicated regex, but it's what Lodash uses (NOTE: `words()` may need to be updated if Lodash has changed their regex since pydash was last updated).

Basically, any function in pydash that needs to break things up on "words" should use the `words()` function. All of the `*_case()` functions are using `words()` under the hood.

bharadwajyarlagadda: My fault. I haven't used the whitespace character. I did this: `''.join(words(text)).lower()`

bharadwajyarlagadda:
@dgilland

pydash/strings.py
dgilland: BTW, I took a quick look at the Lodash regex for words and it's gotten quite a bit more complex and they now have two different regexes that are used for matching ASCII and Unicode separately (i.e. they check if the string has unicode or not and then use one of the two word regexes):

- reUnicodeWord: https://github.com/lodash/lodash/blob/da329eb776a15825c04ffea9fa75ae941ea524af/lodash.js#L257
- reAsciiWord: https://github.com/lodash/lodash/blob/da329eb776a15825c04ffea9fa75ae941ea524af/lodash.js#L161

bharadwajyarlagadda: Will update the regex in the next PR

dgilland: No rush on updating that regex. The other functions are higher priority.

bharadwajyarlagadda:
Good to merge? If so, will squash the commits into one.

dgilland:
Squash away!

bharadwajyarlagadda:
@dgilland Done. Merge away!

PR #70, opened 2016-10-15
Added numerical.subtract() method

bharadwajyarlagadda:


bharadwajyarlagadda:
@dgilland

pydash/numerical.py
```diff
@@ -383,6 +384,30 @@ def std_deviation(array):
...
+        >>> subtract(10, 4)
+        6
+        >>> subtract(-10, 4)
+        -14
+        >>> subtract(4, -10)
+        14
+
+    .. versionadded:: TODO
+    """
+    return minuend - subtrahend
```
dgilland: Take a look at Lodash's "createMathOperation" function which they call for add, subtract, multiply, and divide. Handles a few of the edge cases around inputs.

bharadwajyarlagadda:
@dgilland

pydash/numerical.py
```diff
@@ -383,6 +386,43 @@ def std_deviation(array):
     return math.sqrt(variance(array))
 
 
+def call_math_operator(value1, value2, callback):
```
dgilland: Instead of `callback`, should probably be named `operator` or `operation` or `op`. The name "callback" is too generic while the function being passed in will only be a specific math operation.

pydash/numerical.py
```diff
@@ -383,6 +386,43 @@ def std_deviation(array):
...
+        1.0
+        >>> call_math_operator(4, 6, lambda x, y: x * y)
+        24.0
+        >>> call_math_operator('5', '10', lambda x, y: x + y)
+        15.0
+        >>> call_math_operator('25', '5', lambda x, y: x / y)
+        5.0
+
+    .. versionadded:: TODO
+    """
```
dgilland: Helper functions don't get this level of detail in the docstrings. They're not meant for public use so only really need a brief description. No need for args, returns, or example sections.

pydash/numerical.py
```diff
@@ -483,3 +523,6 @@ def rounder_func(item):
             pass
 
     return result
+
+
+subtract = partial(call_math_operator, callback=operator.sub)
```
dgilland: Le'ts not use a partial here. Can just call `call_math_operator` inside the `subtract` function.

pydash/numerical.py
```diff
@@ -383,6 +386,43 @@ def std_deviation(array):
...
+    # If the user has given any Nonetype values, we assign 0 as the default
+    # value to the values.
+    value1 = 0 if not value1 else value1
+    value2 = 0 if not value2 else value2
+
+    if isinstance(value1, str) or isinstance(value2, str):
+        value1 = float(value1)
+        value2 = float(value2)
+
+    return float(callback(value1, value2))
```
dgilland: No need to cast to `float` here. As long as both values are numbers, it's fine to let the operation return whatever the type would be (e.g. `int + int = int`, `int + float = float`, etc).

pydash/numerical.py
```diff
@@ -383,6 +386,43 @@ def std_deviation(array):
...
+        5.0
+
+    .. versionadded:: TODO
+    """
+    # If the user has given any Nonetype values, we assign 0 as the default
+    # value to the values.
+    value1 = 0 if not value1 else value1
+    value2 = 0 if not value2 else value2
+
+    if isinstance(value1, str) or isinstance(value2, str):
```
dgilland: This comparison won't cover Python 2 since it has the unicode string type. Need to use the `string_types` from `_compat.py`.

However, we should probably instead check if the values are numbers and if not then try to convert them to float since the user may use non-strings or non-numbers that could support conversion to float.

This will tie into whatever policy we decide to implement for handling of non-convertible types. Generally, I think it better if we raise an exception if a non-convertible object is passed in. With that in mind, I think we should do the following:

- If value is `None`, convert to a default value.
- If value is `not pyd.is_number(val)`, then try to convert it to a float but `pass` on any exception.
- If float conversion fails, go ahead with operation and let it throw it's exception (i.e. we don't need to throw it explicitly). This way the exception occurs during the operation and not the attempted float conversion.

pydash/numerical.py
```diff
@@ -383,6 +386,43 @@ def std_deviation(array):
...
+        value1 = float(value1)
+        value2 = float(value2)
```
dgilland: These will fail when one of the values is not convertible to `float`. We should decide whether an invalid type should either be converted to a default value (e.g. `0` or `1`) or whether we should raise a `TypeError` exception.

pydash/numerical.py
```diff
@@ -383,6 +386,43 @@ def std_deviation(array):
...
+        15.0
+        >>> call_math_operator('25', '5', lambda x, y: x / y)
+        5.0
+
+    .. versionadded:: TODO
+    """
+    # If the user has given any Nonetype values, we assign 0 as the default
+    # value to the values.
+    value1 = 0 if not value1 else value1
+    value2 = 0 if not value2 else value2
```
dgilland: The default value should be an argument to this function since `0` won't always be valid (e.g. we shouldn't use `0` for division or multiplication since for division we might get a division-by-zero error and for multiplication we'd get `0` as the return value). Instead, for division/multiplication we'd probably want the default to be `1`.

bharadwajyarlagadda:
@dgilland If good, will squash the commits into one.

pydash/numerical.py
```diff
@@ -383,6 +385,47 @@ def std_deviation(array):
     return math.sqrt(variance(array))
 
 
+def call_math_operator(value1, value2, op):
+    """Return the result of the math operation on the given values.
+
+    .. versionadded:: TODO
```
dgilland: No need for `versionadded` for non-"public"/helper functions.

bharadwajyarlagadda:
@dgilland

pydash/numerical.py
```diff
@@ -383,6 +385,56 @@ def std_deviation(array):
...
+            value2 = float(value2)
+    except Exception:
+        pass
+
+    try:
+        return op(value1, value2)
+    except ZeroDivisionError:
+        # The user can pass in the value2 as 0. When the op is division, then
+        # we return INFINITY.
+        return INFINITY
```
dgilland: I don't think this should be done. If the user divides by zero, then it's an exception. So just do call the `op()` without try/except.

pydash/numerical.py
```diff
@@ -383,6 +385,56 @@ def std_deviation(array):
...
+def call_math_operator(value1, value2, op):
+    """Return the result of the math operation on the given values."""
+    # Note: If the op is division or multiplication, we need to assign 1 as
+    # default value to the values(needs to be passed in from the outer
+    # methods).
+    try:
+        if not pyd.is_number(value1):
+            value1 = float(value1)
+        if not pyd.is_number(value2):
+            value2 = float(value2)
```
dgilland: Each `float()` call should have it's own try/except. It's possible that a value may not be float-able, but may have magic methods for whatever op is being called.

pydash/numerical.py
```diff
@@ -383,6 +385,56 @@ def std_deviation(array):
...
+        -14
+        >>> subtract(2, 0.5)
+        1.5
+
+    .. versionadded:: TODO
+    """
+    # If the user has given any Nonetype values, we assign 0 as the default
+    # value to the values.
+    minuend = 0 if not minuend else minuend
+    subtrahend = 0 if not subtrahend else subtrahend
```
dgilland: The defaults should be set in the `call_math_operator` function so the default value should be an argument to it. Also, we're only assigning a default if a value is `None`.

tests/test_numerical.py
```diff
@@ -156,6 +158,44 @@ def test_std_deviation_aliases(case):
...
+    with pytest.raises(TypeError) as exc:
+        _.subtract(minuend, subtrahend)
+
+    assert error_msg in str(exc)
+
+
+@parametrize('dividend,divisor,expected', [
+    (1, 0, float("inf"))
+])
+def test_divide(dividend, divisor, expected):
```
dgilland: The `test_divide` function should test the `divide` function. Generally, we don't have tests for the helper functions that are not a part of the main pydash API.

pydash/numerical.py
bharadwajyarlagadda: Check out this: https://lodash.com/docs/4.16.4#divide. Lodash is giving a `INFINITY` value for `1/0`.

pydash/numerical.py
bharadwajyarlagadda: I have used the defaults here because, we can assign `1` as default value for `math` and `divide` ops (if the values are `None`)

pydash/numerical.py
dgilland: That's Javascript behavior. Try it with in JS or Node console:

```node
> 1/0
Infinity
```

Python doesn't have this behavior.

pydash/numerical.py
bharadwajyarlagadda: That would be less burden rather than to validate the op and setting the default in `call_math_operator` method. If we want to add the default value in `call_math_operator` method, we can do that too (as it is a inner method used internally in pydash and not a public method).

dgilland: If we decide to change how defaults are set, it would be better to only have a single place that sets the defaults vs. setting the default in every math operation function and having the logic implement in all of those places.

bharadwajyarlagadda:
@dgilland

pydash/numerical.py
```diff
@@ -383,6 +385,52 @@ def std_deviation(array):
     return math.sqrt(variance(array))
 
 
+def call_math_operator(value1, value2, op):
```
dgilland: The default value should be an argument to this function that the main math function will pass in.

tests/test_numerical.py
```diff
@@ -156,6 +158,32 @@ def test_std_deviation_aliases(case):
...
+    assert _.subtract(minuend, subtrahend) == expected
+
+
+@parametrize('minuend,subtrahend,error_msg', [
+    ('abs', 4, 'unsupported operand type'),
+    (4, 'abc', 'unsupported operand type'),
+    ('abs', 'abc', 'unsupported operand type'),
+    ([1, 2, 3, 4], 4, 'unsupported operand type'),
+])
+def test_subtract_error(minuend, subtrahend, error_msg):
```
dgilland: Naming convention used with pydash exception tests is `*_exception`.

tests/test_numerical.py
```diff
@@ -156,6 +158,32 @@ def test_std_deviation_aliases(case):
...
+    with pytest.raises(TypeError) as exc:
+        _.subtract(minuend, subtrahend)
+
+    assert error_msg in str(exc)
```
dgilland: We don't really need this assertion since we're not raising these exceptions ourselves.

pydash/numerical.py
dgilland: Like this:

```python
def call_math_operator(value1, value2, op, default):
```

bharadwajyarlagadda:
@dgilland

pydash/numerical.py
```diff
@@ -383,6 +385,52 @@ def std_deviation(array):
     return math.sqrt(variance(array))
 
 
+def call_math_operator(value1, value2, op, default):
+    """Return the result of the math operation on the given values."""
+    # TODO: Add defaults for math and div operations.
+    if op in (operator.add, operator.sub):
+        value1 = default if not value1 else value1
```
dgilland: Only do this if value is `None`:

```python
if value1 is None:
    value1 = default

if value2 is None:
    value2 = default
```

pydash/numerical.py
```diff
@@ -383,6 +385,52 @@ def std_deviation(array):
     return math.sqrt(variance(array))
 
 
+def call_math_operator(value1, value2, op, default):
+    """Return the result of the math operation on the given values."""
+    # TODO: Add defaults for math and div operations.
+    if op in (operator.add, operator.sub):
```
dgilland: This isn't necessary to check. If we want `default` to be optional, then we'd give it a value of `None` and only use the default is it was `not None`.

bharadwajyarlagadda:
@dgilland

dgilland:
LGTM after squash :+1:

bharadwajyarlagadda:
@dgilland Done.

PR #73, opened 2016-10-17
Added utilities.ntharg() method

bharadwajyarlagadda:


bharadwajyarlagadda:
@dgilland

bharadwajyarlagadda:
@dgilland

pydash/utilities.py
```diff
@@ -393,6 +394,34 @@ def noop(*args, **kargs):  # pylint: disable=unused-argument
     pass
 
 
+def nth_arg(pos):
```
dgilland: This should have a default of `0` and also be used when a non-number is given. If the `pos` is not an integer, then it should be rounded up (so always convert to int but always round up). That is how Lodash's `nthArg` works.

pydash/utilities.py
```diff
@@ -393,6 +394,34 @@ def noop(*args, **kargs):  # pylint: disable=unused-argument
...
+        22
+        >>> func = nth_arg(-1)
+        >>> func(11, 22, 33, 44)
+        44
+
+    .. versionadded:: TODO
+    """
+    def _nth_arg(*args):
+        """Returns argument at the given position :attr:`pos`."""
+        return args[pos]
```
dgilland: This should return `None` if `args` doesn't have an element at `pos`.

pydash/utilities.py
```diff
@@ -393,6 +394,34 @@ def noop(*args, **kargs):  # pylint: disable=unused-argument
...
+        >>> func(11, 22, 33, 44)
+        22
+        >>> func = nth_arg(-1)
+        >>> func(11, 22, 33, 44)
+        44
+
+    .. versionadded:: TODO
+    """
+    def _nth_arg(*args):
+        """Returns argument at the given position :attr:`pos`."""
```
dgilland: If you're going to give the created function a docstring, then may as well replace `pos` in the description with the actual value given. Otherwise, you could skip the docstring. None of the other utility functions set a docstring on returned functions.

tests/test_utilities.py
```diff
@@ -161,6 +161,16 @@ def test_noop(case, expected):
     assert _.noop(*case) == expected
 
 
+@parametrize('args,pos,expected', [
+    ([11, 22, 33, 44], 0, 11),
+    ([11, 22, 33, 44], -1, 44),
+    ([11, 22, 33, 44], -4, 11)
```
dgilland: Should add some test cases for `pos` that is higher/lower than what's given (e.g. 4 and -5) and other cases for default value, string pos, float pos, etc.

bharadwajyarlagadda:
@dgilland

pydash/utilities.py
```diff
@@ -393,6 +394,36 @@ def noop(*args, **kargs):  # pylint: disable=unused-argument
...
+        >>> func(11, 22, 33, 44)
+        22
+        >>> func = nth_arg(-1)
+        >>> func(11, 22, 33, 44)
+        44
+
+    .. versionadded:: TODO
+    """
+    def _nth_arg(*args):
+        if args and (len(args) > pos >= -len(args)):
```
dgilland: You can just do a try/except and catch an IndexError. Or you can use pyd.get() which handles this behavior.

dgilland: Also need to handle the case when the user provides a string where `pos` would be 0 and the case where a non-integer number which would have `pos` be rounded up unconditionally.

dgilland:
@bharadwajyarlagadda ?

pydash/utilities.py
dgilland: If user provides a non-number for pos, then pos should be 0. 

If user provides a non-integer number for pos, then that number should be rounded up to next whole number.

bharadwajyarlagadda: Gotcha

bharadwajyarlagadda:
@dgilland

bharadwajyarlagadda:
@dgilland If good, will squash away.

pydash/utilities.py
```diff
@@ -393,6 +394,36 @@ def noop(*args, **kargs):  # pylint: disable=unused-argument
...
+        22
+        >>> func = nth_arg(-1)
+        >>> func(11, 22, 33, 44)
+        44
+
+    .. versionadded:: TODO
+    """
+    def _nth_arg(*args):
+        try:
+            return pyd.get(args, round(float(pos)))
```
dgilland: Should use `math.ceil` instead of `round`. A float for `pos` should *always* be rounded up.

dgilland: Would prefer something more along the lines of:

```python
try:
    pos = math.ceil(float(pos))
except ValueError:
    pos = 0

return pyd.get(args, pos)
```

That way there's just a single call to `pyd.get()`.

bharadwajyarlagadda: I did the two return statements because it is throwing me an error `UnboundLocalError: local variable 'pos' referenced before assignment`

bharadwajyarlagadda: I guess it is due to the scope of the variable `pos` when I am using it in a inner function's try/except block.

bharadwajyarlagadda: Ok, I actually sorted it out. No worries

bharadwajyarlagadda:
@dgilland

dgilland:
LGTM :+1:

PR #77, opened 2016-10-31
Return default value in get_item when key is missing and cannot be parsed as an int

### LINKED ISSUE #76: ValueError: invalid literal for int() with base 10 in get
Operating System: MacOS X 10.11.5 (El Capitan)
Python versions: 2.7, 3.4, 3.5
Pydash version: 3.4.5
Reproduction steps:

1. In a Python virtualenv with pydash as a pip dependency, open a Python REPL and enter:
```
import pydash
a = { 'a': { 'b': { 'c': { 'd': [] } } } }
pydash.get(a, 'a.b.e.d', [])
``` 
2. You will get a TypeError stating that list indices must be integers, not str, followed by:
```
ValueError: invalid literal for int() with base 10: 'd'
```

It should have returned the default value `[]`

urbnjamesmi1:
Fixes #76 

Handle case in get_item in pydash/helpers.py, where attempt to turn key 
into an int fails.

In this case, raise a KeyError so that it will flow into the
default value logic and ultimately return the default value.


dgilland:
Nice find and thanks for the fix! :+1:

hunt3r:
Nice work @urbnjamesmi1 👍

Issue #78, opened 2016-11-01
When get_item returns the default value, pydash.objects.get continues evaluating the key path


urbnjamesmi1:
If the remaining keys are in the default value, you'll get the value inside the default value supplied to get instead of the default value itself.

Operating System: Mac OS X 10.11.5 (El Capitan)
Steps to reproduce:

1. In a python REPL with pydash installed, type:
```
import pydash

obj = {'one': {'two': {'three': 4}}}
default_value = [{'a': 1}]
value = pydash.get(obj, 'one.four.0.a', default=default_value)
value
```

2. `value` will be 1 instead of `[{'a': 1}]` because while processing the 'four' key, get_item returns the default value and get continues with '0.a'

dgilland:
Awesome work finding this! Really appreciate it! :+1: 

I've released the fix as `v3.4.7`: https://pypi.python.org/pypi/pydash/3.4.7

PR #79, opened 2016-11-01
When pydash.objects.get uses a default value, stop processing keys in

urbnjamesmi1:
the path and just return the default value.

Solves #78 .

### DIFF
--- a/pydash/helpers.py
+++ b/pydash/helpers.py
@@ -21,6 +21,12 @@ class _NoValue(object):
     pass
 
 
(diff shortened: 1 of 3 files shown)


dgilland:
This is superseded by 855f4b63484c527475fc391c02ad09f14eedc521. I decided to go with a different approach.

If you find any issues with 855f4b63484c527475fc391c02ad09f14eedc521, then feel free to open an issue for discussion.

PR #82, opened 2016-11-24
Added predicates.eq() method

bharadwajyarlagadda:


bharadwajyarlagadda:
@dgilland

pydash/predicates.py
```diff
@@ -77,6 +78,34 @@
...
+        >>> eq(None, '')
+        False
+        >>> eq('a', 'a')
+        True
+        >>> eq(1, str(1))
+        False
+
+    .. versionadded:: TODO
+    """
+    return is_equal(value, other)
```
dgilland: Lodash's `eq` function does strict equality comparison: https://github.com/lodash/lodash/blob/1665876a1762d57987246354c4e8d61034aa1f20/lodash.js#L11198

So you should do the comparison with `is` instead of using `is_equal()`.

bharadwajyarlagadda: I guess it should be `==`.

bharadwajyarlagadda: I think using `is` can be unpredictable.

dgilland: JavaScript '===' is the same as 'is' in Python. So 'eq' should check whether the two objects being compared are the same object.

bharadwajyarlagadda:
@dgilland

PR #83, opened 2016-11-27
Added methods is_equal_with, is_set, nth and flip

bharadwajyarlagadda:


bharadwajyarlagadda:
@dgilland

pydash/functions.py
```diff
@@ -210,6 +211,32 @@ def callback(item):
...
+        >>> flipped(1, 2, 3, 4)
+        (4, 3, 2, 1)
+        >>> flipped = flip(lambda args: [i * 2 for i in args])
+        >>> flipped(1, 2, 3, 4)
+        [8, 6, 4, 2]
+
+    .. versionadded:: TODO
+    """
+    def _flip(*args):
+        return func(args)[::-1]
```
dgilland: This is supposed to reverse the arguments given, not the result. And it's not supposed to convert `*args` into a tuple.

Should be defined more like:

```python
def _flip(*args, **kargs):
    return func(*reversed(args), **kargs)
```

Also, instead of defining as an inline function, should go ahead and define it as a class similar to how all the other functions in this module are defined (for consistency).

pydash/predicates.py
```diff
@@ -535,6 +537,9 @@ def is_equal(value, other, callback=None):
     return equal
 
 
+is_equal_with = is_equal
```
dgilland: `is_equal_with` and `is_equal` should be separate functions. Lodash split the original functionality of `is_equal` into `is_equal` and `is_equal_with`. Basically, `is_equal_with` should work just like `is_equal` does previously where it accepts a callback. Then `is_equal` should become a new function that doesn't allow a callback and instead just calls `is_equal_with`.

tests/test_arrays.py
```diff
@@ -307,6 +307,15 @@ def test_mapcat(case, expected):
     assert _.mapcat(*case) == expected
 
 
+@parametrize('case,pos,expected', [
+    ([11, 22, 33], 2, 33),
+    ([11, 22, 33], 0, 11),
+    ([11, 22, 33], -1, 33)
+])
```
dgilland: Should add test cases for out-of-range indexes.

bharadwajyarlagadda:
@dgilland

pydash/functions.py
```diff
@@ -210,6 +211,32 @@ def callback(item):
...
+        >>> flipped = flip(lambda *args: args)
+        >>> flipped(1, 2, 3, 4)
+        (4, 3, 2, 1)
+        >>> flipped = flip(lambda *args: [i * 2 for i in args])
+        >>> flipped(1, 2, 3, 4)
+        [8, 6, 4, 2]
+
+    .. versionadded:: TODO
+    """
+    def _flip(*args, **kargs):
```
dgilland: Per my previous comment, let's implement this as a class like the other functions in this submodule.

bharadwajyarlagadda:
@dgilland

pydash/functions.py
```diff
@@ -210,6 +211,33 @@ def callback(item):
         return pyd.some(obj, callback)
 
 
+class flip(object):
```
dgilland: This is not quite how these classes and functions are defined for this submodule. All functions have a corresponding class that implements the function wrapper logic. So there should be a class Flip() and a function flip() where the flip() function returns and instance of the Flip() class. See other function//class pairs for examples.

bharadwajyarlagadda:
@dgilland

dgilland:
:+1:

PR #84, opened 2016-12-02
Added cond(), default_to(), stub_list(), stub_dict(), stub_false(), stub_string(), stub_true() methods

bharadwajyarlagadda:


bharadwajyarlagadda:
@dgilland

pydash/utilities.py
```diff
@@ -680,6 +748,86 @@ def result(obj, key, default=None):
     return ret
 
 
+def stub_array():
```
dgilland: Since we're using `stub_dict` instead of `stub_object`, we should use `stub_list` instead of `stub_array`.

pydash/utilities.py
```diff
@@ -85,6 +92,42 @@ def attempt(func, *args, **kargs):
...
+    Example:
+
+        >>> func = cond([matches({'a': 1}), constant('matches A')])
+        >>> func({'a': 1, 'b': 2})
+        'matches A'
+
+    .. versionadded:: TODO
+    """
+    for pair in pairs:
+        if len(pair) != 2:
```
bharadwajyarlagadda: Should I default pairs to anything in `except` block?

pydash/utilities.py
```diff
@@ -136,6 +179,31 @@ def deep_property(path):
...
+    Example:
+
+        >>> default_to(1, 10)
+        1
+        >>> default_to(None, 10)
+        10
+
+    .. versionadded:: TODO
+    """
+    return default_value if not value else value
```
bharadwajyarlagadda: Just 
```python
return default_value if not value
```??

File: pydash/utilities.py (hunk 2 above)
Comment (dgilland) [inline pydash/utilities.py]: No, you should still raise the `ValueError` but what I'm saying is that `len(pair)` may fail so you need to handle that too in order to ensure that the `ValueError` is raised vs. whatever error might be thrown when you do `len(pair)`.

An alternative for this could be:

```python
try:
    assert len(pair) == 2
except Exception:
    raise ValueError(...)

if not all(map(callable, pair)):
    raise TypeError(...)
```

pydash/utilities.py
dgilland: Something like this:

```python
return default_value if value is None else value
```

bharadwajyarlagadda:
@dgilland

Issue #66, opened 2016-10-11
Add LoDash functions


bharadwajyarlagadda:
## Arrays
- [x] `difference_by`
- [x] `difference_with`
- [x] `flatten_depth`
- [x] `from_pairs`
- [x] `intersection_by`
- [x] `intersection_with`
- [x] `nth`
- [x] `pull_all`
- [x] `pull_all_by`
- [x] `pull_all_with`
- [x] `sorted_index_by`
- [x] `sorted_index_of`
- [x] `sorted_last_index_by`
- [x] `sorted_last_index_of`
- [x] `sorted_uniq`
- [x] `sorted_uniq_by`
- [x] `union_by`
- [x] `union_with`
- [x] `uniq_by`
- [x] `uniq_with`
- [x] `xor_by`
- [x] `xor_with`
- [x] `zip_object_deep`

## Collections
- [x] `flat_map`
- [x] `flat_map_deep`
- [x] `flat_map_depth`
- [x] `invoke_map`
- [x] `sample_size`

## Objects
- [x] `assign_with`
- [x] `clone_deep_with`
- [x] `clone_with`
- [x] `invert_by`
- [x] `merge_with`
- [x] `omit_by`
- [x] `pick_by`
- [x] `set_with`
- [x] `to_integer`
- [x] `unset`
- [x] `update`
- [x] `update_with`

## Predicates
- [x] `eq`
- [x] `is_equal_with`
- [x] `is_match_with`
- [x] `is_set`

## Strings
- [x] `lower_case`
- [x] `lower_first`
- [x] `upper_case`
- [x] `upper_first`
- [x] `to_lower`
- [x] `to_upper`

## Numerical
- [x] `clamp`
- [x] `mean_by`
- [x] `max_by`
- [x] `min_by`
- [x] `sum_by`
- [x] `subtract`
- [x] `divide`
- [x] `multiply`

## Functions
- [x] `flip`
- [x] `unary`

## Utilities
- [x] `cond`
- [x] `conforms`
- [x] `conforms_to`
- [x] `default_to`
- [x] `nth_arg`
- [x] `over`
- [x] `over_every`
- [x] `over_some`
- [x] `range_right`
- [x] `stub_array` (or maybe `stub_list`?)
- [x] `stub_object` (or maybe `stub_dict`?)
- [x] `stub_string`
- [x] `stub_false`
- [x] `stub_true`
- [x] `to_path`

dgilland:
All of these functions have been added now.

PR #85, opened 2016-12-24
getargspec -> getfullargspec in python 3

tgriesser:
Currently seeing a `ValueError` if a callback specifies type annotations

### DIFF
--- a/pydash/functions.py
+++ b/pydash/functions.py
@@ -6,11 +6,10 @@
 
 from __future__ import absolute_import
 
-import inspect
 import time
 
 import pydash as pyd
-from ._compat import _range
+from ._compat import _range, getfullargspec
 
 
 __all__ = (
@@ -134,7 +133,7 @@ class Curry(object):
     """Wrap a function in a curry context."""
     def __init__(self, func, arity, args=None, kargs=None):
         self.func = func
-        self.arity = (len(inspect.getargspec(func).args) if arity is None
+        self.arity = (len(getfullargspec(func).args) if arity is None
                       else arity)
         self.args = () if args is None else args
         self.kargs = {} if kargs is None else kargs
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-
 
-from copy import deepcopy
-
 import pytest
 
 from pydash._compat import iteritems
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -1,9 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import pytest
-
 from pydash._compat import iteritems
-
 
 # pytest.mark is a generator so create alias for convenience
 parametrize = pytest.mark.parametrize
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -15,3 +15,4 @@ Contributors
 - Klaus Sevensleeper, k7sleeper@gmail.com, `k7sleeper@github <https://github.com/k7sleeper>`_
 - Bharadwaj Yarlagadda, yarlagaddabharadwaj@gmail.com, `bharadwajyarlagadda@github <https://github.com/bharadwajyarlagadda>`_
 - Michael James, `urbnjamesmi1 <https://github.com/urbnjamesmi1>`_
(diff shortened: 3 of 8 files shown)


pydash/functions.py
```diff
@@ -10,7 +10,7 @@
 import time
 
 import pydash as pyd
-from ._compat import _range
+from ._compat import _range, PY3
```
dgilland: Instead of doing the `PY3` check in each module, I'd instead define `getfullargspec` in `_compat.py` and import that function where needed:

```python
# _compat.py
if PY3:
    getfullargspec = inspect.getfullargspec
else:
    getfullargspec = inspect.getargspec
```

tests/fixtures.py
```diff
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 
-from copy import deepcopy
-
 import pytest
-
+from sys import version_info
 from pydash._compat import iteritems
 
+if version_info[0] > 2:
```
dgilland: A better way to handle this would be to use `pytest.skipif`: http://doc.pytest.org/en/latest/skipping.html#id1

```python
import pytest

from pydash._compat import PY3

@pytest.mark.skipif(not PY3, reason='requires python3')
def test_annotated_callback():
    ...
```

I'd also avoid putting actual test functions in `fixtures.py` and instead put in a `test_*.py` file. Maybe `test_annotations.py`?

tasks.py
```diff
@@ -16,7 +16,7 @@
     'no-value-for-parameter',
     'method-hidden'
 ])
-TEST_TARGETS = ' '.join([PACKAGE_NAME, 'tests'])
+TEST_TARGETS = ' '.join([PACKAGE_NAME, 'tests/test_*'])
```
dgilland: Can probably leave this change off once the `tests/_compat.py` is moved to a `test_*.py` file.

tests/fixtures.py
tgriesser: Python 2.7 actually raises a `SyntaxError` for a file containing type annotations, which is why I don't believe the `pytest.mark.skipif` will work and instead went with the dynamic import & changed the test targets below. Wasn't sure the best way to approach this.

pydash/functions.py
tgriesser: Good call! Fixed.

tests/fixtures.py
dgilland: Would this approach work: http://stackoverflow.com/a/35230792

where you define the annotations in a comment?

dgilland: Or does that not modify the function spec which wouldn't really test that annotations are handled correctly?

dgilland: I'd like to avoid having to modify the pytest arguments in `tasks.py` (i.e. by changing test targets from `'tests'` to `'tests/test_*'`) and relying on conditionally importing the test function into a non-"test_" prefixed file (`fixtures.py`).

Since the annotations test is very minor, I'd prefer to add it in a way that requires the fewest changes even if the solution is a hack (which it will be :smile:):

```python
# test_annotations.py

import pytest

import pydash as _
from pydash._compat import PY3


if PY3:
    # Hack around the fact that we can't define this kind of function in
    # Python 2 so have to rely on conditionally creating it using exec()
    # to avoid syntax errors while still having this test case covered for
    # Python 3.
    exec('def typed_function(row: int, index: int, col: list): return row + 1')
else:
    typed_function = None


@pytest.mark.skipif(not PY3, reason='test requires Python 3 annotations')
def test_annotated_callback():
    assert _.map_([1, 2], typed_function) == [2, 3]
```

It's not pretty but I think it'll do for now.

tasks.py
tgriesser: Fixed

tests/fixtures.py
tgriesser: Great idea - thanks!

tgriesser:
Updated with `test_annotations.py` as suggested & rebased

dgilland:
Awesome work! :+1:

dgilland:
I'll work on releasing this for the 3.x series some time this week.

If you're interested in helping out for the 4.0 release, that'd be much appreciated! Feel free to "claim" some of the functions from #66 by adding a comment so we don't end up working on the same things.

dgilland:
FYI, this patch has been released as `v3.4.8`: https://pypi.python.org/pypi/pydash/3.4.8

PR #87, opened 2017-01-28
Added arrays and utilities methods

bharadwajyarlagadda:


bharadwajyarlagadda:
@dgilland

pydash/arrays.py
```diff
@@ -1813,6 +1912,9 @@ def iterunique(array, comparator=None, iteratee=None):
             seen.append(cmp_item)
 
 
+iterunion = iterunique
```
dgilland: There isn't really a reason to alias. Can just use `iterunique` directly where needed.

pydash/arrays.py
```diff
@@ -1440,7 +1489,57 @@ def union(*arrays):
...
+    Example:
+
+        >>> union_by([1, 2, 3], [2, 3, 4], callback=lambda x: x % 2)
+        [1, 2]
+        >>> union_by([1, 2, 3], [2, 3, 4], callback=lambda x: x % 9)
+        [1, 2, 3, 4]
+
+    .. versionadded:: TODO
+    """
+    return list(iterunion(flatten(arrays), iteratee=callback))
```
dgilland: Since `union()` calls `uniq()`, may as well do the same here but with `return uniq_by(flatten(arrays), callback=callback)`.

pydash/arrays.py
```diff
@@ -1440,7 +1489,57 @@ def union(*arrays):
...
+
+        >>> comparator = lambda a, b: (a % 2) == (b % 2)
+        >>> union_with([1, 2, 3], [2, 3, 4], callback=comparator)
+        [1, 2]
+        >>> union_with([1, 2, 3], [2, 3, 4])
+        [1, 2, 3, 4]
+
+    .. versionadded:: TODO
+    """
+    return list(iterunion(flatten(arrays), comparator=callback))
```
dgilland: Since `union()` calls `uniq()`, may as well do the same here but with `return uniq_with(flatten(arrays), callback=callback)`.

pydash/arrays.py
bharadwajyarlagadda: Well using `iterunique` method for a `union` method makes me confusing. I would think, how come "they are using unique method inside union methods". You can say the functionality is the same and why a different name. But, imagine a person coming and seeing the code - he/she would think the same as above.

dgilland: My other comment says to use the uniq_by and uniq_with functions for union_by/with respectively so there's no reason for this function. I'm not liking to alias helper functions that are only used internally. That just adds unnecessary extras that serve little purpose.

bharadwajyarlagadda:
@dgilland

pydash/arrays.py
```diff
@@ -1424,7 +1466,7 @@ def take_while(array, callback=None):
     return array[:n]
 
 
-def union(*arrays):
+def union(arrays):
```
dgilland: The function signature shouldn't change.

pydash/arrays.py
```diff
@@ -1435,14 +1477,64 @@ def union(*arrays):
...
-        >>> union([1, 2, 3], [2, 3, 4], [3, 4, 5])
+        >>> union([[1, 2, 3], [2, 3, 4], [3, 4, 5]])
         [1, 2, 3, 4, 5]
 
     .. versionadded:: 1.0.0
     """
     return uniq(flatten(arrays))
 
 
+def union_by(arrays, callback=None):
```
dgilland: Let's mirror `union` by having `*arrays` as the argument instead of only `arrays`.

pydash/arrays.py
```diff
@@ -1435,14 +1477,64 @@ def union(*arrays):
...
+        [1, 2]
+        >>> union_by([[1, 2, 3], [2, 3, 4]], callback=lambda x: x % 9)
+        [1, 2, 3, 4]
+
+    .. versionadded:: TODO
+    """
+    return uniq_by(flatten(arrays), callback=callback)
+
+
+def union_with(arrays, callback=None):
```
dgilland: Same here: `*arrays`

pydash/arrays.py
dgilland: Actually, take a look at `intersection`, `intersection_by`, and `intersection_with`. We should model the function signatures the same for the `union*` functions. There's also special handling for passing in the callback as a positional argument that can be used in `union_by` and `union_with`.

bharadwajyarlagadda:
@dgilland

pydash/arrays.py
```diff
@@ -1443,6 +1487,58 @@ def union(*arrays):
     return uniq(flatten(arrays))
 
 
+def union_by(*arrays, **kargs):
```
dgilland: Take a look at the intersection() functions. We want to do the same function signature for the union() functions.

bharadwajyarlagadda: Something like this:
```python
def union_by(array, *others, **kargs):
    return uniq_by(flatten(array + others), callback=kargs.get('callback'))
```

Or you want to change the implementation too??

bharadwajyarlagadda: Well not exactly the same. I need to add this too:

```python
if isinstance(others, tuple):
    others = [others]
```

dgilland: Look at intersection_by(), for example, and just duplicate the handling of others and kargs. Basically the callback should be passable as either a positional arg or a keyword arg.

dgilland:
@bharadwajyarlagadda Any progress on this?

bharadwajyarlagadda:
@dgilland

bharadwajyarlagadda:
@dgilland

PR #88, opened 2017-02-09
Added arrays.sorted_index_of() and arrays.pull_all() methods

bharadwajyarlagadda:


bharadwajyarlagadda:
@dgilland

pydash/arrays.py
```diff
@@ -1322,6 +1348,31 @@ def sorted_index(array, value, callback=None):
...
+        >>> sorted_index_of([10, 5, 7, 3], 3)
+        0
+        >>> sorted_index_of([10, 10, 5, 7, 3], 10)
+        3
+        >>> sorted_index_of([10, 5, 7, 3], 11)
+        -1
+
+    .. versionadded:: TODO
+    """
+    return sorted(array).index(value) if value in array else -1
```
dgilland: This isn't quite what this function is supposed to do as the output in your examples doesn't match the corresponding function calls in Lodash's `.sortedIndexOf`:

```javascript
> var lo = require('lodash');
> lo.sortedIndexOf([10, 5, 7, 3], 3)
-1
> lo.sortedIndexOf([10, 10, 5, 7, 3], 10)
-1
> lo.sortedIndexOf([10, 5, 7, 3], 11)
-1
```

If you look at the Lodash source, then you can see that Python equivalent is basically:

```python
index = sorted_index(array, value)
if index < len(array) and array[index] == value:
    return index
else:
    return -1
```

* https://github.com/lodash/lodash/blob/master/sortedIndexOf.js
* https://github.com/lodash/lodash/blob/master/sortedIndex.js
* https://github.com/lodash/lodash/blob/master/.internal/baseSortedIndex.js

bharadwajyarlagadda: The documenatation in lodash is quite confusing to me. It says `performs a binary search on a sorted array`. So is the array already sorted or being sorted as part of the functionality?

I can see that the functionality is around an array which is already sorted. The array is not sorted as part of the functionality.

bharadwajyarlagadda:
@dgilland

bharadwajyarlagadda:
@dgilland

Issue #92, opened 2017-04-04
no replacement for deep_pluck with list argument


efenka:
With the upgrade to version 4, there seems to be no equal replacement for `deep_pluck` with a list as argument.The recommended `.map` call has a different behavior.

Version 3.4.8:

    >>> py_.deep_pluck([{1: {2: {3: 4}}}], [1, 2])
    [{3: 4}]

Version 4.0.0:

    >>> py_.map([{1: {2: {3: 4}}}], [1, 2])
    [False]

The only workaround i found was:

    >>> py_.map([{1: {2: {3: 4}}}], py_.property_deep([1, 2]))
    [{3: 4}]

dgilland:
You're right. `map_` doesn't support the same functionality as `deep_pluck` did (which was an oversight on my part).

I've brought the functionality of `deep_pluck` back in `4.0.1` but renamed it to `pluck`: https://pypi.python.org/pypi/pydash/4.0.1

As another consequence, I've also removed `property_deep` and made ` property_` work like `property_deep` did.

PR #94, opened 2017-04-06
add example for intersection of a single list

efenka:
This is to foresee #92 in the future via doctests.


dgilland:
Thanks :+1:

PR #97, opened 2017-05-31
Use standard replace instead of regex

shaunpatterson:
pydash.get can be quite slow.  

`unescaped_path_key` was using `reg_exp_js_replace` to replace `\\` and `\.` in keys.  This seems like overkill.  Instead, use standard string replace. 

In my tests this seems to speed up `_.get` by a factor of 8-10x


dgilland:
:+1:

dgilland:
FYI, this has been released in `v4.0.4`: https://pypi.python.org/pypi/pydash/4.0.4

PR #98, opened 2017-06-01
modify `iteratee` to support multiple arguments

yunfan:


dgilland:
Do you have some use cases for this functionality? Just curious how you might use it.

yunfan:
@dgilland  yes i have put that in the doc test
```python
       >>> iteratee(('a', 'b'))({'a': 1, 'b': 2, 'c': 3})
        [1, 2]
```

dgilland:
This functionality basically mimics `pydash.at`:

```python
>>> pydash.at({'a': 1, 'b': 2, 'c': 3}, 'a', 'b')
[1, 2]
```

So we could add a new `properties()` function:

```python
# pydash/utilities.py

def properties(*paths):
    return lambda obj: [getter(obj)
                        for getter in (pyd.property_(path) for path in paths)]
```

Then `iteratee()` could use `properties()` when a tuple is passed in (like in this PR) to support deep property mapping:

```python
>>> iteratee(('a', 'b', ['c', 'd', 'e']))({'a': 1, 'b': 2, 'c': {'d': {'e': 3}}})
[1, 2, 3]
```

Later, I can modify `at()` to use `properties()` under the hood for deep property support there too.

yunfan:
i am ok with your alternative implement. just waiting that.

PR #106, opened 2017-11-01
simplify usage of potentially single values

efenka:
I often have the problem that i don't know if a value given by a user is a list or a single value. My API allows to give for example a list of database ids to work on or just a single id. For this reason i have to check if the given value is already a list / tuple and if not, transform it before i can apply further pydash functions.

For this reason it would be nice to have a pydash function that wraps single values in a list. Especially for chaining.

My issue:
```
>>> py_(1).sum().value()
0
```

Simplification using listify:
```
>>> listify(1)
[1]
```

```
>>> listify([1])
[1]
```

```
>>> listify({1: 2, 3: 4})
[1, 3]
```

```
>>> listify(a for a in [1, 2, 3])
[1, 2, 3]
```

Chained:
```
>>> py_(1).listify().sum().value()
1
```

```
>>> py_([1, 2, 3]).listify().sum().value()
6
```

### DIFF
--- a/pydash/__init__.py
+++ b/pydash/__init__.py
@@ -46,6 +46,7 @@ from .arrays import (
     intersperse,
     last,
     last_index_of,
+    listify,
     mapcat,
     nth,
     pop,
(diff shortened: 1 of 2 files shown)


efenka:
Sorry i did not see that i had an old commit in this pull request. I will close it and open another clean pull requests.

PR #107, opened 2017-11-01
simplify usage of potentially single values

efenka:
I often have the problem that i don't know if a value given by a user is a list or a single value. My API allows to give for example a list of database ids to work on or just a single id. For this reason i have to check if the given value is already a list / tuple and if not, transform it before i can apply further pydash functions.

For this reason it would be nice to have a pydash function that wraps single values in a list. Especially for chaining.

My issue:
```
>>> py_(1).sum().value()
0
```

Simplification using listify:
```
>>> listify(1)
[1]
```

```
>>> listify([1])
[1]
```

```
>>> listify({1: 2, 3: 4})
[1, 3]
```

```
>>> listify(a for a in [1, 2, 3])
[1, 2, 3]
```

Chained:
```
>>> py_(1).listify().sum().value()
1
```

```
>>> py_([1, 2, 3]).listify().sum().value()
6
```


dgilland:
This method already exists as [to_list](https://github.com/dgilland/pydash/blob/fc6a52770c972507addd048389e84fce9c8efcc3/pydash/collections.py#L1052). It doesn't have the same functionality as what's in this PR though.

efenka:
I think it is somehow a similar idea, but for my use case it's not sufficient. Unfortunately `py_.to_list(1)` returns `TypeError: 'int' object is not iterable` but not `[1]`.

dgilland:
I'd prefer to modify `to_list` with the functionality you've implemented here.

efenka:
Sure :) Should i do it and send you another pull request or want you to modify it yourself?

dgilland:
Feel free to submit a new PR with the functionality moved to `to_list`.

efenka:
Hm, looks my implementation has some compatibility differences between python2 and python3. I will check fix next within the next days.

pydash/collections.py
```diff
@@ -1049,31 +1049,51 @@ def sort_by(collection, iteratee=None, reverse=False):
     return sorted(collection, key=pyd.iteratee(iteratee), reverse=reverse)
 
 
-def to_list(collection):
```
dgilland: Keep the variable name as is.

pydash/collections.py
```diff
@@ -1049,31 +1049,51 @@ def sort_by(collection, iteratee=None, reverse=False):
...
+
     """
-    if isinstance(collection, dict):
-        ret = collection.values()
-    else:
-        ret = list(collection)
 
-    return ret
+    if isinstance(item, list):
+        return item
```
dgilland: This function needs to return a new object and not the same object.

pydash/collections.py
```diff
@@ -1049,31 +1049,51 @@ def sort_by(collection, iteratee=None, reverse=False):
...
 
+    elif isinstance(item, dict):
+        return item.values()
+
+    elif hasattr(item, '__iter__'):
+        return list(item)
+
+    else:
+        return [item]
```
dgilland: The reason this is failing on Travis is because Python 2.7 `str` doesn't have an `__iter__` method. So with these changes, on Python 3, `to_list('abc') == ['a', 'b', 'c']` but on Python 2 it's `to_list('abc') == ['abc']`.

The current behavior is that `to_list('abc')` should be `['a', 'b', 'c']`. Therefore, this could be simplified to:

```python
if isinstance(collection, dict):
    ret = collection.values()
else:
    try:
        ret = list(collection)
    except Exception:
        ret = [collection]

return ret
```

This next question then is whether you are proposing that `to_list('abc')` should return `['abc']`? Changing this behavior would break backwards compatibility so cannot do that on the `v4` line.

efenka: Yes i saw it. Unfortunately my local test environment offers only python3.6, what somehow works not so well with tox for me. Anyway. 

I'm proposing that converting a single string to a list shot wrap this string into a list and not split it by character. We could add an parameter for this, but actually the python2 behavior feels more natural for me.

dgilland: In that case, adding a parameter for making strings return as a wrapped list would be fine. Next major release could make that the default behavior potentially.

efenka:
Unfortunately I had no time in the past days to continue the work here. So with some delay, please check the latest changes i just pushed. 

I also found some differences in byte type handling of the old implementation - caused by the fact that bytes is just an alias for str in python2. I tried to make this more reliable, but maybe you don't like the solution.

dgilland:
Thanks again for the PR!

PR #111, opened 2018-01-15
allow to use to_list as decorator

efenka:
I often write functions that collect data as list and then somehow return them. For this reason it's mostly required to create an internal variable to collect the data and return that at the end. To safe these repeating lines, I modified the to_list function to also act as decorator, so i can use ``yield`` to create lists.

    >>> @py_.to_list
    >>> def foo():
    >>>     yield "b"
    >>>     yield "a"
    >>>     yield "r"
    >>>     if "something" is True:
    >>>         yield "s"
    >>>
    >>> foo()
    ['b', 'a', 'r']

Actually i'm not sure if you want functionality like this in pydash. But let's give it a try 🙃. I also added a test to the memoize function, where i use the same technique to call it.

### DIFF
--- a/pydash/objects.py
+++ b/pydash/objects.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import
 
 import copy
 from functools import partial
+from functools import wraps
 import math
 import re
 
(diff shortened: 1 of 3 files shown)


dgilland:
Think I'll punt on the decorator implementation for `to_list`. Not something I want to add (would start to clutter up the function implementations).

Issue #100, opened 2017-07-10
FP Implementation


patrick1011:
Hi Folks,

Great library, been using it over the last month or two and enjoying it.  I was just wondering if anyone has ever looked into porting over FP?  I quite like wrapping my functions in `flow` which really necessitates 'auto-curried iteratee-first data-last' functions to quote the FP blurb.  I've taken a look at how FP is written and would be up for implementing something similar in pydash if you think it's a sensible idea!  If so I'll get something together a make a PR!

Cheers,

Patrick

dgilland:
I'm not necessarily opposed to the idea in theory, but I'd like to hear what kind of design you would propose to port `lodash/fp` since it seems like it would require a large amount of work to do (at least from an initial glance at https://github.com/lodash/lodash/wiki/FP-Guide).

patrick1011:
Have you seen https://github.com/lodash/lodash/tree/4.17.4-npm/fp ?  https://github.com/lodash/lodash/blob/4.17.4-npm/fp/_baseConvert.js is where most of the action takes place.  

I think the base conversion would generally be a variation on the below and could cover upwards of half of the library.
```
def pydash_fp_func(collection):
    def wrapped(predicate):
        pydash_func(collection, predicate)
    return wrapped
```
Special cases would of course likely crop up.  And the tests would need to be ported over too.

dgilland:
Ok, so basically there would be a set of wrapper functions to convert pydash functions to fp-style functions. Sounds reasonable enough. Let's keep the changes in a feature branch like `feature/fp`.

patrick1011:
Yeah.  Nice one - I'll get something together for you to take a look at.

patrick1011:
Just to let you know I'm still working on this - progress has been slow because I've been ill the last week.

dgilland:
Any progress on this?

impguard:
:(

Unfortunate. It makes a lot more sense as a builtin for the library (from one perspective), since both would always be in sync. Instead of having one library trail the other.

Though I can see why the decision was made.

PR #114, opened 2018-02-07
fp implementation

altaurog:
Here’s an implementation of FP version (#100)

Pretty much everything revolves around a `convert` function which applies a series of wrapper functions for currying, argument reordering, immutability, and argument capping.  It also does some fancy stuff with the docstring.

Setting the functions up is pretty simple, but adjusting the tests is a bit time consuming.  I mostly just copied the test cases directly, but not all of the original test cases work with the fp variants because almost none of the fp functions have optional arguments.

I completed just `pydash.fp.arrays` and would like to get some feedback.  If this looks acceptable I can continue on the rest, time permitting, and perhaps others will volunteer to assist.  I would like to make some improvements in the docstring manipulation (and may have to adjust it in any case as more functions are added) but what we have here is certainly respectable.

Here are [the generated docs](http://static.aryehleib.com/pydash/fp.html)

I initially used `pydash.curry` and `pydash.rearg` but ultimately ended up writing a separate implementation for each.  I reimplemented `curry` in order to support a placeholder and `rearg` in order to support varargs.  With `pydash.rearg` arguments are still optional.  That is not relevant for the fp variants, so my implementation is less flexible in that regard:

```
In [19]: h = pydash.rearg(lambda *args: args, 1, 2, 0)
In [20]: h(1, 2)
Out[20]: (2, 1)
In [21]: g = convert.rearg_ex([1, 2, 0])(lambda *args: args)
In [22]: g(1, 2)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-22-dfd039199aeb> in <module>()
----> 1 g(1, 2)

~/git/pydash/pydash/fp/convert.py in <lambda>(*args)
     90     """
     91     def transform(f):
---> 92         return lambda *args: f(*getargs(order, interpose, args))
     93     return transform
     94 

~/git/pydash/pydash/fp/convert.py in getargs(order, interpose, args)
     96 def getargs(order, interpose, args):
     97     count = len(order)
---> 98     base = operator.itemgetter(*order)(args)
     99     if interpose:
    100         index = order.index(max(order)) + 1

IndexError: tuple index out of range
```

As a side point, I find it striking that these (and many other of the functions in `pydash.functions`) are implemented using classes.  Perhaps that accounts for the slightly better performance with my implementations:

```
In [1]: f = lambda a, b, c: (a, b, c)

In [2]: %timeit g = convert.curry_ex(3)(f)
550 ns ± 4.18 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit h = pydash.curry(f, 3)
668 ns ± 4.02 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: g = convert.curry_ex(3)(f)

In [5]: h = pydash.curry(f, 3)

In [6]: %timeit g(1)(2)(3)
3.14 µs ± 27.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: %timeit h(1)(2)(3)
5.13 µs ± 37.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [8]: %timeit g = convert.rearg_ex([1, 2, 0])(f)
511 ns ± 9.36 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [9]: %timeit h = pydash.rearg(f, 1, 2, 0)
3.46 µs ± 20.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [10]: g = convert.rearg_ex([1, 2, 0])(f)

In [11]: %timeit g(1, 2, 3)
681 ns ± 3.08 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [12]: h = pydash.rearg(f, 1, 2, 0)

In [13]: %timeit h(1, 2, 3)
2.1 µs ± 28.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

I also added a function [`pydash.arrays.repack`](http://static.aryehleib.com/pydash/api.html#pydash.arrays.repack) to provide support in python 2.7 for extended iterable unpacking, which I used in the tests.


dgilland:
Hi @altaurog! This looks great! Awesome work and a good start to the FP module! :+1: 

I haven't had a chance to deep dive into the code, but I have done a high-level overview. So far it seems like a good approach to solving the problem. A few initial comments:

1. RE: fp docstrings: I wonder what the performance impact will be on the initial import since the docstrings are manipulated at import time. Mainly concerned with how that might affect CLI tools that use pydash and trying not to have a noticeable delay if `pydash.fp` is used. One idea to mitigate this (if it becomes an issue), would be to move to a code-generator approach that builds the `fp.<module>` files from spec files. Output would look similar to the manual version except that the docstrings would be rendered and assigned like `<function>.__doc__ = '''...'''`. That would also eliminate any requirement for the library user to install `astor` or any other dev-related lib and provide consistent docstrings.
2. RE: repack and 2.7 support: Is continued Python 2.7 support something you are interested in? I'd actually like to drop support for Python 2.7, possibly in the next major release of pydash. If not having to support 2.7 in `fp` would make things easier to develop/test, then I'd be fine with making `fp` Python >= 3 only.
3. RE: classes in pydash.functions: This was something I did way back in the initial release of pydash and a pattern that I continued within that module. I wasn't as comfortable with function wrappers/closures back then. Converting these to pure functions would be fine to do. Again, something to include in the next major pydash release.

I'm going to try to carve out some time this week to review the code in more detail, but so far things are off to a good start. Thank you for taking the time to work on this; I appreciate it!

altaurog:
Thanks @dgilland!

I would really like to reorganize the docstring conversion as it is a bit haphazard.  It would definitely be worth profiling that to see what the performance impact is, and certainly the fp docstrings could be generated once at build time (i.e. before the package is published).  My schedule is a bit hard to predict, but when I next have a little chunk of time I will try to get to that.

altaurog:
I did some primitive profiling and indeed the docstring conversion is much more expensive than the rest of the conversion, on my machine 0.024s / 0.0013s.  So here is a code-generation approach, as you have suggested.  I also fixed the doctests and a number of other issues with the docstrings.

Regarding python2.7 support, it may be worth continuing.  Here are download stats for the last 30 days:
```
$ pypinfo --pip --markdown --percent pydash pyversion
```

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  49.25% |          5,651 |
| 3.6            |  32.35% |          3,711 |
| 3.5            |  12.34% |          1,416 |
| 3.4            |   6.02% |            691 |
| 3.7            |   0.03% |              4 |

altaurog:
@dgilland fair ’nuf.  I am exploring ideas for an independent library as an alternative to this PR.  In any case, thanks for getting back to this; it is good to get definite closure in any case.

PR #120, opened 2018-08-26
Fixes broken `lodash-contrib` link

andres-lowrie:
I'm guessing this new path https://github.com/node4good/lodash-contrib is the correct repo, the old link was broken


dgilland:
Thanks! :+1:

PR #123, opened 2018-11-16
Change quickstart imports/usage to be more consistent

AetherUnbound:
Just a minor thing I noticed about the Quickstart guide - `flatten` is imported into the namespace whereas all the other functions are used against the `pydash` namespace. This PR changes them all to be consistent on this page.


dgilland:
Thanks for this!

PR #131, opened 2019-11-08
Fixed bug where getattr is used on Mappings and Sequence in Python 3.5+

jwilson8767:
I found a case where objects in Python 3.5+ which use generic collections (Sequence, Mapping, not sure about others) were having `getattr` applied to them, creating an issue where `get(mymapping, 'values')` would return the mapping values instead of the key 'values'.

This PR resolves the issue mentioned above. Versions of python < 3.5 should not be affected. I was able to run tests successfully, but I am missing several interpreter versions on my machine so I ask that you re-test if there isn't a CI setup. I did not add a unit test case to confirm this behavior because I'm not sure how to have python version specific tests for tox.

P.S. I noticed that requirements.txt and requirements-dev.txt appear to have been removed/clobbered at some point in the past, but are still mentioned in `contributing.rst`.

### DIFF
--- a/src/pydash/helpers.py
+++ b/src/pydash/helpers.py
@@ -11,15 +11,23 @@ import warnings
 import pydash as pyd
 from ._compat import Iterable, PY2, iteritems, getfullargspec, string_types
 
+_attr_ignored_types = (dict, list)
+if not PY2:
+    try:
+        from typing import Sequence, Mapping
+        _attr_ignored_types = (dict, list, Sequence, Mapping)
+    except Exception:
+        pass
+
 
 class _NoValue(object):
-    """Represents an unset value. Used to differeniate between an explicit
+    """Represents an unset value. Used to differentiate between an explicit
     ``None`` and an unset value.
     """
     pass
 
 
-#: Singleton object that differeniates between an explicit ``None`` value and
+#: Singleton object that differentiates between an explicit ``None`` value and
 #: an unset value.
 NoValue = _NoValue()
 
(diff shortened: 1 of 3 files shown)


src/pydash/helpers.py
```diff
@@ -11,15 +11,23 @@
 import pydash as pyd
 from ._compat import Iterable, PY2, iteritems, getfullargspec, string_types
 
+_attr_ignored_types = (dict, list)
+if not PY2:
+    try:
+        from typing import Sequence, Mapping
```
dgilland: Instead of importing from `typing` module (typically that's used for type hinting and not runtime type checking), we can import from `collections.abc` (Python 3) or `collections` (Python 2) and then can use the same set of types for both versions. We can put the conditional import in `pydash._compat` and then import `Sequence` and `Mapping` from there. Then, we'd only need `(Sequence, Mapping)` since `dict/list` instances are types of those.

src/pydash/helpers.py
```diff
@@ -142,9 +150,10 @@ def base_get(obj, key, default=NoValue):
     except Exception:
         pass
 
-    if not isinstance(obj, (dict, list)):
+    if not isinstance(obj, _attr_ignored_types) or hasattr(obj, '_fields'):
```
dgilland: For `namedtuple` detection, we can use `or (isinstance(obj, tuple) and hasattr(obj, "_fields"))` to help minimize false-positives since it's possible some other custom classes might define a `_fields` attribute but not be a namedtuple.

src/pydash/helpers.py
jwilson8767: That sounds perfect to me.

src/pydash/helpers.py
jwilson8767: Good catch!

Issue #136, opened 2020-02-13
crash in `collections.order_by()` if None in one of ordering keys


vlad0337187:
Hello.

Found very unpleasant bug when worked with sorting.
Hope, this can be fixed soon =)

(attached code which should fix it below)

1. if None in one of ordering keys

```python
import pydash
collections_0_4 = [{'a': 1, 'b': 3}, {'a': 1, 'b': None}]
result_0_4 = pydash.collections.order_by(collections_0_4, ['a', 'b'])
```

error: 
```
Traceback (most recent call last):
  File "/home/vlad/Temp/issue_pydash/issue_reproduce.py", line 13, in <module>
    result_0_3 = pydash.collections.order_by(collections_0_3, ['a', 'b'])    # reproduced
  File "/home/vlad/.local/share/virtualenvs/issue_pydash-TqXHV2Jm/lib/python3.6/site-packages/pydash/collections.py", line 675, in order_by
    return sorted(collection, key=cmp_to_key(comparison), reverse=reverse)
  File "/home/vlad/.local/share/virtualenvs/issue_pydash-TqXHV2Jm/lib/python3.6/site-packages/pydash/collections.py", line 669, in comparison
    result = _cmp(func(left), func(right))
  File "/home/vlad/.local/share/virtualenvs/issue_pydash-TqXHV2Jm/lib/python3.6/site-packages/pydash/_compat.py", line 86, in _cmp
    def _cmp(a, b): return (a > b) - (a < b)
TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
```

2. if one of ordering keys missing

```python
import pydash
collections_6 = [{'a': 0, 'b': 1}, {'a': 0, 'b': 2, 'c': 3}, {'a': 0, 'b': 3, 'c': 4}]
result_6 = pydash.collections.order_by(collections_6, ['a', 'c'])
```

error: 
```
$ pipenv run python '/home/vlad/Temp/issue_pydash/issue_reproduce.py' 
another issue not reproduced if this message is visible
Traceback (most recent call last):
  File "/home/vlad/Temp/issue_pydash/issue_reproduce.py", line 35, in <module>
    result_6 = pydash.collections.order_by(collections_6, ['a', 'c'])    # reproduced
  File "/home/vlad/.local/share/virtualenvs/issue_pydash-TqXHV2Jm/lib/python3.6/site-packages/pydash/collections.py", line 675, in order_by
[...]

PR #137, opened 2020-03-04
Fixes issue #136

elijose55:
Fixes issue #136
Changed function **_cmp** to accept NoneType arguments. If one of the arguments is None it considers it as the smaller one. If both are None then the function returns 0, as if they were equal. Passed all tests in Tox.

### DIFF
--- a/src/pydash/_compat.py
+++ b/src/pydash/_compat.py
@@ -84,13 +84,13 @@ else:
     izip = zip
 
     def _cmp(a, b):
-            if (a is None and b is None):
-                return 0
-            elif (a is None):
-                return -1
-            elif(b is None):
-                return 1
-            return (a > b) - (a < b)
+        if (a is None and b is None):
+            return 0
+        elif (a is None):
+            return -1
+        elif(b is None):
+            return 1
+        return (a > b) - (a < b)
 
 
 builtins = dict((value, key) for key, value in iteritems(_builtins.__dict__)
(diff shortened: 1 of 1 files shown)


src/pydash/_compat.py
```diff
@@ -83,7 +83,14 @@ def iteritems(d): return iter(d.items())
...
     izip = zip
 
-    def _cmp(a, b): return (a > b) - (a < b)
+    def _cmp(a, b):
+            if (a == None and b == None):
+                return 0
+            elif (a == None):
+                return -1
+            elif(b == None):
+                return 1
```
dgilland: Use `is None` for comparison to `None`.

dgilland:
Just a minor change in the `None` comparison and then this is good to merge.

Thanks!

src/pydash/_compat.py
elijose55: Okay! Thanks for the help!

src/pydash/_compat.py
```diff
@@ -83,7 +83,14 @@ def iteritems(d): return iter(d.items())
...
 
-    def _cmp(a, b): return (a > b) - (a < b)
+    def _cmp(a, b):
+            if (a is None and b is None):
+                return 0
+            elif (a is None):
+                return -1
+            elif(b is None):
+                return 1
+            return (a > b) - (a < b)
```
slallum: indentation fix :)
```suggestion
        if (a is None and b is None):
            return 0
        elif (a is None):
            return -1
        elif(b is None):
            return 1
        return (a > b) - (a < b)
```

Issue #108, opened 2017-11-16
pick() does not support @property properties 


m3talstorm:
**Example code:**

```
import pydash

class Test(object):

    @property
    def blah(self):
        return 5


test = Test()

print pydash.pick(test, 'blah')


class Test(object):

    def __init__(self):

        self.blah = 5


test = Test()

print pydash.pick(test, 'blah')

```
**Output:**

```
{}
{'blah': 5}
```

**Version:**

4.0.3

Is this intended? Haven't tested other descriptors.

m3talstorm:
Looks like it uses the iterator helper function from [here](https://github.com/dgilland/pydash/blob/develop/pydash/helpers.py#L86) which in turn uses [this](https://github.com/dgilland/pydash/blob/develop/pydash/helpers.py#L97)

`getattr(obj, '__dict__', {})`

To access the object's properties, __dict__ does not return @property 's

dgilland:
Excluding class properties was an explicit intention; they just weren't considered when support for iterating over class attributes was added. The original idea was to have a fairly lightweight way to exclude class methods.

But I'm not opposed to extending iteration to include class properties as well.

PR #138, opened 2020-03-09
Fix/consider property annotation

hugosoftdev:
This modification is useful to enable that objects derivated from classes with a property declared as a method+decorator also can be "picked" by the .pick() function.

Fixing the issue #108 .

### DIFF
--- a/src/pydash/helpers.py
+++ b/src/pydash/helpers.py
@@ -111,7 +111,7 @@ def iterator(obj):
     elif isinstance(obj, Iterable):
         return enumerate(obj)
     else:
-        return iteritems(getattr(obj, '__dict__', {}))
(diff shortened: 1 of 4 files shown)


src/pydash/helpers.py
```diff
@@ -111,7 +111,7 @@ def iterator(obj):
     elif isinstance(obj, Iterable):
         return enumerate(obj)
     else:
-        return iteritems(getattr(obj, '__dict__', {}))
+        return iteritems(dict((attribute, getattr(obj, attribute)) for attribute in dir(obj) if not attribute.startswith('__')))
```
dgilland: This will end up selecting attributes that are meant to be excluded like methods. The intention around using `obj.__dict__` was to only return attributes set on a class. It's not very sophisticated and misses things like properties, but that was intentional.

For those cases, one can pass in a custom iteratee that would select the desired attributes from the class.

Issue #129, opened 2019-07-26
pick_by has different behavior with iteratee=None compared to lodash


tailsu:
In Lodash the default iteratee is `_.identity` (for all fooBy functions) so `_.pickBy(x)` returns a dictionary with only those keys that have truthy values. I can't figure out what `pydash.pick_by(x)` does. It seems to always return an empty dictionary.

dgilland:
That is indeed what it does which doesn't seem very useful. So `pydash.pick_by(obj)` should be modified to return a copy of the object instead of an empty dictionary since it's pretty worthless otherwise.

tailsu:
I'd prefer if pick_by (and all foo_by) functions retain the lodash's semantics when called without an explicit iteratee, namely use _.identity.

dgilland:
Most of the other *_by() functions have the expected behavior for iteratee=None but for whatever reason pick_by was missed which should definitely be fixed.

elijose55:
Can I work on this issue?

dgilland:
@elijose55 Go for it!

elijose55:
Should I change the behavior for iteratee=None of the pick_by function to return a copy of the object?

dgilland:
I think the implementation just needs to have something like the following in it:

```python
if iteratee is None:
    iteratee = pyd.identity
    argcount = 1
```

So that when no iteratee is given, it will use the `pyd.identity` as the iteratee.

elijose55:
I tried this implementation and the function returns a copy of the object when the iteratee=None. Thats what is expected right? And one test failed with this:
```sh
tests/test_objects.py::test_pick_by[case2-expected2] FAILED
```

```
case = ([1, 2, 3],), expected = {}

    @parametrize('case,expected', [
        (({'a': 1, 'b': 2, 'c': 3}, ['a', 'b']), {'a': 1, 'b': 2}),
        (({'a': 1, 'b': 2, 'c': 3}, lambda value, key: key in ['a']),
         {'a': 1}),
        (([1, 2, 3],), {}),
        (([1, 2, 3], [0]), {0: 1}),
        ((fixtures.Object(a=1, b=2, c=3), 'a'), {'a': 1}),
        ((fixtures.ItemsObject({'a': 1, 'b': 2, 'c': 3}), 'a'), {'a': 1}),
        ((fixtures.IteritemsObject({'a': 1, 'b': 2, 'c': 3}), 'a'), {'a': 1}),
    ])
[...]

PR #139, opened 2020-03-11
Fixes issue #129

elijose55:
Fixed pick_by function behavior: 
- When the **iteratee** is **None** the function uses **pyd.identity** as the **iteratee**. 
- Updated the test related to that function, with the case **iteratee = None**, and changed the expected return to a copy of the object.


dgilland:
Thanks!

PR #151, opened 2020-10-20
Implementing `default_to_any`

gonzalonaveira:
Adding the method `default_to_any`  to allow having multiple default values to fallback


gonzalonaveira:
Hi @dgilland,
Review if you can please, 
Thanks!

CHANGELOG.rst
```diff
@@ -3,6 +3,11 @@
 Changelog
 =========
 
+v4.8.1 (2020-10-20)
```
dgilland, line 6: I generally don't add release version info until I'm ready to release. Can just leave this section heading out.

src/pydash/__version__.py
```diff
@@ -2,4 +2,4 @@
 """Project version information.
 """
 
-__version__ = '4.8.0'
+__version__ = '4.8.1'
```
dgilland, line 5: Similar to the CHANGELOG comment, this is only bumped once a release is getting cut. So no need to change this now.

src/pydash/utilities.py
```diff
@@ -262,6 +262,34 @@ def default_to(value, default_value):
...
+
+        >>> default_to_any(1, 10, 20)
+        1
+        >>> default_to_any(None, 10, 20)
+        10
+        >>> default_to_any(None, None, 20)
+        20
+
+
+    .. versionadded:: 4.8.1
```
dgilland, line 288: Can just mark the version as `x.x.x` for now.

tests/test_utilities.py
```diff
@@ -135,6 +135,17 @@ def test_default_to(case, expected):
     assert _.default_to(*case) == expected
 
 
+@parametrize('case,expected', [
+    (([1, 10, 20]), 1),
+    (([None, 10, 20]), 10),
+    (([None, None, 20]), 20),
+    (([None, [1, 2], [3, 4]]), [1, 2]),
+    (([None, None, [3, 4]]), [3, 4])
+])
```
dgilland, line 144: Can add a case for all `None` values as input just to verify `None` is returned in that case.

src/pydash/utilities.py
```diff
@@ -262,6 +262,34 @@ def default_to(value, default_value):
...
+    """
+    return pyd.reduce_(default_values, pyd.default_to, value)
```
dgilland, line 290: I think instead of having `default_to_any` call `default_to`, we can just make `default_to` a special case of `default_to_any` and thus have `default_to` call `default_to_any` like:

```python
def default_to_any(value, *default_values):
    values = (value,) + default_values
    for val in values:
        if val is not None:
           return val

def default_to(value, default_value):
    return default_to_any(value, default_value)
```

dgilland (changes requested):
Thanks for the PR! Just a few minor comments.

gonzalonaveira:
Thanks for the review and the comments

dgilland (approved):

PR #152, opened 2020-10-21
pydash.set_ support key with dot to work as pydash.get

zhaowb:
`pydash.set_({}, r'a.b\.c.c', 1)`

### DIFF
--- a/src/pydash/objects.py
+++ b/src/pydash/objects.py
@@ -1670,6 +1670,8 @@ def update_with(obj, path, updater, customizer=None):
         base_set(target, key, path_obj, allow_override=False)
 
         try:
+            if isinstance(key, str):
+                key = key.replace('.', r'\.')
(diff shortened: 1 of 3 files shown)


src/pydash/objects.py
```diff
@@ -1670,6 +1670,8 @@ def update_with(obj, path, updater, customizer=None):
         base_set(target, key, path_obj, allow_override=False)
 
         try:
+            if isinstance(key, str):
+                key = key.replace('.', r'\.')
             target = get(target, key)
```
dgilland, line 1675: I think this could just be changed to `base_get(target, key, default=None)` instead of having to escape the dots.

Some test cases for this would be good too.

dgilland (changes requested):

dgilland (approved):

PR #166, opened 2021-08-31
refactor: :recycle: replace some op with stdlib functions && add some types

wuyudi:


wuyudi:
- [x] add
- [x] subtract
- [x] multiply
- [x] divide
- [x] eq
- [x] gt
- [x] gte
- [x] lt
- [x] lte

wuyudi:
About tox:

only this https://github.com/houndci/hound/issues/1769#issuecomment-572677506, I choose to ignore it.

```
src/pydash/arrays.py:479: [BLK100] Black would make changes.
src/pydash/predicates.py:750: [BLK100] Black would make changes.
src/pydash/collections.py:181: [BLK100] Black would make changes.
src/pydash/utilities.py:141: [BLK100] Black would make changes.
src/pydash/strings.py:232: [BLK100] Black would make changes.
```

and

src/pydash/strings.py:262: [E501] line too long (102 > 100 characters)

the other tests all passed.

dgilland:
While I appreciate the effort here, I won't be able to accept this PR. Aside from the style changes (which are not necessary since the `black` tool takes care of formatting <- not running `black` leads to those `[BLK100] Black would make changes.` errors), the other changes are not needed and might actually break backwards compatibility.

Several functions are being changed in subtle ways that are not equivalent (e.g. `return list(itermap(collection, iteratee))` changed to `return map(iteratee, collection)` is not the same thing. The `itermap()` function does more than the builtin `map` and the return type of `map` is a map object and not a list).

Replacing pydash functions with stdlib, while probably something that could have been done originally, is potentially risky since they may not be exactly equal (i.e. there's a non-zero risk of breaking things for little gain; it's legacy at this point unfortunately). Unless there is a bug in one of the functions? That would warrant a fix if so. Also, just ripping out the functions and replacing them as stdlib imports would change the docstrings for them. Wrapper functions would be a better solution so the docstrings would stay but, again, there's a risk to changing the implementation at this point and "if it ain't broke, don't fix it".

Type hints would be a nice addition but is not something I can commit to maintaining due to the level of effort involved (i.e. would need to be all-or-nothing thing; less than 100% type-hinting wouldn't be desirable).

In the future, it would be a good idea to first open an issue with a proposal before making large changes (that are not bug fix related) so that a discussion can be had first. This project is in maintenance mode so larger features/changes have to be considered carefully.

PR #177, opened 2023-02-04
Typehints

DeviousStoat:


dgilland:
Just a heads up that I'm planning to review this soon. But from just skimming it, this looks like some really awesome work, @DeviousStoat! Thanks for a taking the time to work on this!

DeviousStoat:
Awesome! I did some tests on my own but there are so many functions and different ways to call them so I am pretty sure there will be some typing issues. However if it comes to be merged I would be happy to keep an eye on reported issues and make fixes as they come

src/pydash/arrays.py
```diff
@@ -90,8 +96,13 @@
...
 )
 
+T = t.TypeVar("T")
+T2 = t.TypeVar("T2")
+SequenceT = t.TypeVar("SequenceT", bound=t.Sequence)
+MutableSequenceT = t.TypeVar("MutableSequenceT", bound=t.MutableSequence)
+
 
-def chunk(array, size=1):
+def chunk(array: t.List[T], size: int = 1) -> t.List[t.List[T]]:
```
dgilland, line 105: It's been some time since I've done much with Python typing, but I'm not sure about requiring the arguments to be an iterable of the same type (e.g. `foo(t.List[T]) -> t.List[t.List[T]]`). Unless specified, there usually aren't restrictions/requirements that the iterable contain a single type. Or am I misunderstanding how `[T]` works where it's ok if, e.g., a list has a mixture of strings and integers? Or would `chunk([1, 'a'])` fail this type hint and we need something like less restrictive like `foo(t.Iterable) -> t.Iterable[t.Iterable]`?

src/pydash/arrays.py
```diff
@@ -1151,11 +1394,19 @@ def push(array, *items):
         Removed alias ``append``.
     """
     for item in items:
-        array.append(item)
-    return array
+        t.cast(t.List[t.Union[T, T2]], array).append(item)
```
dgilland, line 1397: Is there a way to separate the casting from the variable modifications? Like do the casting and then do the append as separate statements? I don't necessarily mind the casting if that's necessary, but I think mixing it with the actual function logic could be confusing.

Also, what is the issue that requires casting in the first place? Is this more because of the usage of `T` and `T2` in the return? See my above comment about whether we need to actually drop iterable-item type enforcement in order to support mixed type iterables.

src/pydash/collections.py
```diff
@@ -1059,6 +2258,24 @@ def some(collection, predicate=None):
...
+
+@t.overload
+def sort_by(
+    collection: t.Iterable[T],
+    iteratee: t.Union[t.Callable[[T], t.Any], IterateeObjT, None] = None,
+    reverse: bool = False,
+) -> t.List[T]:
+    ...
+
+
```
dgilland, line 2278: All these overloads are contributing quite a bit to total line length. Are they all necessary? Wondering if being less strict about objects containing all of the same type would eliminate the need for some of these overloads?

src/pydash/numerical.py
```diff
@@ -40,10 +48,25 @@
     "zscore",
 )
 
+T = t.TypeVar("T")
+T2 = t.TypeVar("T2")
+T3 = t.TypeVar("T3")
+NumT = t.TypeVar("NumT", int, float)
```
dgilland, line 54: Should `Decimal` be added?

Is this different than `NumberT` or `NumberNoDecimalT`?

src/pydash/objects.py
```diff
@@ -535,7 +905,7 @@ def get(obj, path, default=None):
     return obj
 
 
-def has(obj, path):
+def has(obj: t.Iterable, path: t.Union[str, int, t.List[t.Union[str, int]]]) -> bool:
```
dgilland, line 908: `path` should be able to be `path: t.Union[t.Hashable, t.List[t.Hashable]]`.

Similarly for any other `path` argument that's used as dictionary keys.

src/pydash/utilities.py
```diff
@@ -487,23 +622,25 @@ def memoize(func, resolver=None):
...
+    def memoized(*args: P.args, **kwargs: P.kwargs):
         if resolver:
             key = resolver(*args, **kwargs)
         else:
             key = f"{args}{kwargs}"
 
-        if key not in memoized.cache:
-            memoized.cache[key] = func(*args, **kwargs)
+        if key not in t.cast(MemoizedFunc, memoized).cache:
+            t.cast(MemoizedFunc, memoized).cache[key] = func(*args, **kwargs)
```
dgilland, line 632: Here's another example of the casting which I think obscures what's happening. Would definitely prefer separation of casting from logic. And if that's not possible, then I'd prefer not casting with whatever is needed to make the type checker happy (i.e. don't want to sacrifice readability for 100% type correctness within the internals of the functions).

dgilland (changes requested):
First pass review. Overall, looking great! Hoping we can cut down on total file size increase with those overloads by being less restrictive with the type requirements. Less restriction overall would also be good since there's a lot of flexibility builtin into the functions regarding what's valid to pass in.

src/pydash/arrays.py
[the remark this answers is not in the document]
DeviousStoat: I had some troubles using less restrictive types on some occasions, but it seems this `chunk` function is definitely an example that could be typed as `(t.Sequence[T], int) -> t.List[t.Sequence[T]])`.
I will look over and try to find other cases like this one and fix them.

DeviousStoat: Yes a type variable (`T` here) can be a union of types.
So if we take your example when passing `[1, 'a']` to the `chunk` function `T` would be equal to `Union[int, str]` so the full computed type of the function in this case would be `(t.List[Union[int, str]], int) -> t.List[t.List[Union[int, str]]]`

```py
# type_test.py
import typing as t
import pydash as _


my_list: t.List[t.Union[int, str]] = [1, 'a']

x = _.chunk(my_list)

reveal_type(x)
```
```bash
$ mypy type_test.py
type_test.py:9:13: note: Revealed type is "builtins.list[builtins.list[Union[builtins.int, builtins.str]]]"
Success: no issues found in 1 source file
```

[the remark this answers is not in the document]
dgilland: List would probably be needed for cases where the function is using list-specific methods like append, extend, etc.

src/pydash/arrays.py
DeviousStoat: The issue here is that even though python allows mixing types in a list, the type system rightfully tells you some thing is wrong if you append a different type to a list declared of a certain type.

```python
# type_test.py
import typing as t
import pydash as _


my_list: t.List[int] = [1]

my_list.append("a")
```
```bash
$ mypy type_test.py
type_test.py:7:16: error: Argument 1 to "append" of "list" has incompatible type "str"; expected "int"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

But for this function I was thinking more flexibility would be nice. The way I typed it, you can pass a list of integers and a string element and the result will be a list of `Union[int, str]` without any error like `append` would report. This means we have to cast the array to bypass the error appending another type would report.

Another way to right this function could be:
```python
    new_array = t.cast(t.List[t.Union[T, T2]], array)

    for item in items:
        new_array.append(item)
    return new_array
```
We would need another intermediate list because just assigning the casted list would report a wrong assignment error.

Let me know if you prefer this approach

src/pydash/collections.py
DeviousStoat: The main idea of most of these overloads is to separate mappings from iterables. IMO this is needed if we want to type the iteratees properly. `Iterable` is not a `Mapping` and `Mapping` is an `Iterable` but it is not generic on the value of the mapping.

```python
@t.overload
def sort_by(
    collection: t.Mapping[t.Any, T2],
    iteratee: t.Union[t.Callable[[T2], t.Any], IterateeObjT, None] = None,
    reverse: bool = False,
) -> t.List[T2]:
    ...


@t.overload
def sort_by(
    collection: t.Iterable[T],
    iteratee: t.Union[t.Callable[[T], t.Any], IterateeObjT, None] = None,
    reverse: bool = False,
) -> t.List[T]:
    ...
```
So the idea here is that without overload we have no way to say this function accepts any iterable and the iteratee's first argument (if it is a callable) will be the value of the mapping if we use a mapping as `collection`

This is because pydash passes the value of the mapping as first argument to the iteratee which is nice but it makes typing a bit trickier.

src/pydash/arrays.py
DeviousStoat: Another solution would be to ignore type checking too:
```python
    for item in items:
        array.append(item)  # type: ignore
    return array  # type: ignore
```

src/pydash/objects.py
DeviousStoat: This sounds nice. Also made me realize that having a type alias for this would be nice as it is used in several places.

So I added a `PathT` type alias with the hashable union you suggested and used it where possible

src/pydash/arrays.py
dgilland: I think I'd be in favor of just ignoring the type checking for cases where casting would be required. I'm more concerned that the input/output typing is correct and less concerned about the internals being 100% on typing if doing so means there's more complexity to get it "right".

src/pydash/collections.py
dgilland: I was only able to skim the overloads on my first pass through so I was mostly concerned with how many there were.

Looking at this particular example of `sort_by`, could that be simplified to:

```python
def sort_by(
    collection: t.Union[t.Iterable[T], t.Mapping[t.Any, T]],
    iteratee: t.Union[t.Callable[[T], t.Any], IterateeObjT, None] = None,
    reverse: bool = False,
) -> t.List[T]:
    ...
```

If so, are there other functions where we could potentially use `t.Union` to reduce the number of overloads in cases where the other arguments are common between them?

tests/pytest_mypy_testing/test_functions.py
```diff
@@ -0,0 +1,286 @@
...
+@pytest.mark.mypy_testing
+def test_mypy_spread() -> None:
+    def greet_people(*people: str) -> str:
+        return 'Hello ' + ', '.join(people) + '!'
+
+    greet = _.spread(greet_people)
+    reveal_type(greet)  # R: pydash.functions.Spread[builtins.str]
+    reveal_type(greet(['Mike', 'Don', 'Leo']))  # R: builtins.str
+
+@pytest.mark.mypy_testing
```
povilasb, line 259: Awesome!

src/pydash/collections.py
DeviousStoat: Unfortunately this doesn't really work because `Iterable[T]` and `Mapping[t.Any, T]` overlap.
`pyright` seems to pick the less specific, so in this case if we pass a dict, the `Iterable[T]` will be picked and `T` will be the keys' type instead of the values' type.
`mypy` rejects this type altogether, passing a dict, we get an incompatible type error, not exactly sure why.

When I first started typing this library I used a generic type alias for the iteratee type which reduced overload amount a lot. If we pick `drop_while` as a simple example. Current type is:
```python
@t.overload
def drop_while(array: t.List[T], predicate: t.Callable[[T, int, t.List[T]], t.Any]) -> t.List[T]:
    ...


@t.overload
def drop_while(array: t.List[T], predicate: t.Callable[[T, int], t.Any]) -> t.List[T]:
    ...


@t.overload
def drop_while(array: t.List[T], predicate: t.Callable[[T], t.Any]) -> t.List[T]:
    ...


@t.overload
def drop_while(array: t.List[T], predicate: None = None) -> t.List[T]:
    ...
```

And with my generic type alias idea I had:
```python
F = t.Union[
    t.Callable[[T, T2, t.List[T]], T3],
    t.Callable[[T, T2], T3],
    t.Callable[[T], T3],
    None,
]

def drop_while(array: t.List[T], predicate: F[T, int, t.Any] = None):
    ...
```

Which is really nice, you can reduce a lot of overload, and it works for list and dict, depending on the generic values you pass. Here `int` as second because it is a list, for mapping it would be:

```python
F = t.Union[
    t.Callable[[T, T2, t.Dict[T2, T]], T3],
    t.Callable[[T, T2], T3],
    t.Callable[[T], T3],
    None,
]

def some_func_accepting_dict(d: t.Mapping[T2, T], predicate: F[T, T2, t.Any] = None):
    ...
```

This is a bit different than the `sort_by` example above as you still need to have overloads for `Iterable` and `Mapping` but it combines all the different predicate/iteratee overloads in one single overload.

While this works perfectly fine with `pyright`, it doesn't work at all with `mypy`, it just doesn't get it somehow and everything is typed as `Any`. I opened an issue in `mypy` about this when I realized: https://github.com/python/mypy/issues/14498

dgilland: Does mypy make any guarantees around which type it uses from a Union when an arg matches more than one? Like use the first matched from the Union list or anything like that? So if the more specific match like Mapping was first in the Union would that one be used over any other types in the Union? Or is behavior undefined for this situation and can't be relied on?

DeviousStoat: More than undefined behaviour, it doesn't even accept it if both types in the union overlaps it seems.
Small test:
```python
# test_type.py
import typing as t

T = t.TypeVar("T")

def func(x: t.Union[t.Mapping[t.Any, T], t.Iterable[T]]) -> t.List[T]:
    return []

d: t.Dict[int, str] = {5: "hello"}

x = func(d)
reveal_type(x)
```
While `pyright` accepts this but matches the wider type (`Iterable`), we get revealed type `x: List[int]`
`mypy` doesn't even accept it:
```bash
$ mypy test_type.py
test_type.py:14: error: Need type annotation for "x" (hint: "x: List[<type>] = ...")  [var-annotated]
test_type.py:14: error: Argument 1 to "func" has incompatible type "Dict[int, str]"; expected "Union[Mapping[Any, <nothing>], Iterable[<nothing>]]"  [arg-type]
test_type.py:15: note: Revealed type is "builtins.list[Any]"
Found 2 errors in 1 file (checked 1 source file)
```

What seems to work is to restrict the iterable type:
```python
def sort_by(
    collection: t.Union[t.List[T], t.Mapping[t.Any, T]],
    iteratee: t.Union[t.Callable[[T], t.Any], IterateeObjT, None] = None,
    reverse: bool = False,
) -> t.List[T]:
    ...
```

DeviousStoat:
@dgilland if you dislike having a lot of typing overhead in the code (the big amount of overloads).
another solution could be to provide [stub files](https://mypy.readthedocs.io/en/stable/stubs.html) instead.
So the `.py` source files would be unchanged and we would have the types in separate `.pyi` files that would be shipped with the library. I am not too familiar with this approach but I can look into it. From your comments it seems like this approach would maybe suit you better

src/pydash/collections.py
dgilland: So potentially unions of List/Mapping or Sequence/Mapping might work? How much would that potentially reduce the number of overloads needed across functions? If it's significant then that might be something to compromise on. Otherwise, may just bite the bullet and have the overloads to get a little better on the typing.

dgilland:
I don't want stub files. I'm ok with having the overloads where necessary if the alternative is bad typing/too permissive typing. Just want to be sure that the overloads that get added can't be reduced any further. But if we have to add 1k+ lines in some files to get the typing right and useful, then so be it.

src/pydash/collections.py
DeviousStoat: I don't think this would reduce much. This would only be possible in functions that don't accept the full iteratee as implemented in the combination of `iteriteratee` and `callit`. The main thing that requires all these overloads is the combinations of functions accepting `Iterable` or `Mapping` and the different number of arguments the `predicate`/`iteratee` can be called with, as I showed in my last comment.
`sort_by` is an example that don't accept the full arguments iteratee, so it has only 2 overloads and using List/Mapping or Sequence/Mapping would reduce to only 1 overload (eg. no overload needed).
But most functions are typed like this `reject` for example:

```python
@t.overload
def reject(
    collection: t.Mapping[T, T2],
    predicate: t.Union[t.Callable[[T2, T, t.Dict[T, T2]], t.Any], IterateeObjT, None] = None,
) -> t.List[T2]:
    ...


@t.overload
def reject(
    collection: t.Mapping[T, T2],
    predicate: t.Union[t.Callable[[T2, T], t.Any], IterateeObjT, None] = None,
) -> t.List[T2]:
    ...


@t.overload
def reject(
    collection: t.Mapping[t.Any, T2],
    predicate: t.Union[t.Callable[[T2], t.Any], IterateeObjT, None] = None,
) -> t.List[T2]:
    ...


@t.overload
def reject(
    collection: t.Iterable[T],
    predicate: t.Union[t.Callable[[T, int, t.List[T]], t.Any], IterateeObjT, None] = None,
) -> t.List[T]:
    ...


@t.overload
def reject(
    collection: t.Iterable[T],
    predicate: t.Union[t.Callable[[T, int], t.Any], IterateeObjT, None] = None,
) -> t.List[T]:
    ...


@t.overload
def reject(
    collection: t.Iterable[T],
    predicate: t.Union[t.Callable[[T], t.Any], IterateeObjT, None] = None,
) -> t.List[T]:
    ...
```

And since in most of the overloads the signature of the `predicate` changes depending on the `collection` type, I think we can only remove 1 overload:
```python
@t.overload
def reject(
    collection: t.Mapping[T, T2],
    predicate: t.Union[t.Callable[[T2, T, t.Dict[T, T2]], t.Any], IterateeObjT, None] = None,
) -> t.List[T2]:
    ...


@t.overload
def reject(
    collection: t.Mapping[T, T2],
    predicate: t.Union[t.Callable[[T2, T], t.Any], IterateeObjT, None] = None,
) -> t.List[T2]:
    ...


@t.overload
def reject(
    collection: t.Iterable[T],
    predicate: t.Union[t.Callable[[T, int, t.List[T]], t.Any], IterateeObjT, None] = None,
) -> t.List[T]:
    ...


@t.overload
def reject(
    collection: t.Iterable[T],
    predicate: t.Union[t.Callable[[T, int], t.Any], IterateeObjT, None] = None,
) -> t.List[T]:
    ...


@t.overload
def reject(
    collection: t.Union[t.Sequence[T], t.Mapping[t.Any, T]],
    predicate: t.Union[t.Callable[[T], t.Any], IterateeObjT, None] = None,
) -> t.List[T]:
    ...
```

dgilland: Alright, so sounds like we need the overloads as they are. The only other thing I can think of to help reduce the overall line count would be to collapse the overload functions defs to a single line and have the formatter ignore those lines. But I'm not sure much that would hurt readability of the types since some of them might be very difficult to grok if the function def and its typed arguments are on a single line. But for now, I think leaving it as-is will be fine.

src/pydash/numerical.py
DeviousStoat: Yes this is different, `NumberT` is only a `Union` of types, this `NumT` gets bound to one of the types in the signature of a function. Small example:
```python
import typing as t

NumberT = t.Union[int, float]
NumT = t.TypeVar("NumT", int, float)

def add(x: NumberT, y: NumberT) -> NumberT:
    return x + y

def add_generic(x: NumT, y: NumT) -> NumT:
    return x + y
    
reveal_type(add(5, 5))  # Revealed type is "Union[builtins.int, builtins.float]"
reveal_type(add_generic(5, 5))  # Revealed type is "builtins.int"
```
So here `NumberT` is always fixed to a union of types. But `NumT` gets bound to the detected type of `x` and `y` hence the return type is also `int`.

However this is not needed in most of the cases I used it in. I removed it and retyped `clamp` which I think I didn't type well.

scripts/mypy_doctests_generator.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def main(path: Path) -> str:
+    imports = "import pytest\n\n"
+    imports += f"import pydash as _\n\n\n"
+
+    return imports + "\n\n".join(map(generate_test_function, docstrings(path)))
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
DeviousStoat: I added a `generate_mypy_test` task. But be aware that this is not perfect, and only generates boilerplate ready to be filled from doctests. But it probably saved me quite a bit of time

src/pydash/arrays.py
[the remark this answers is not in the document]
DeviousStoat: I found quite a few functions where I used `List` but could actually use `Iterable` or `Sequence`, replaced them

dgilland (approved):
Absolutely awesome PR! Thanks a ton for this!

PR #181, opened 2023-03-01
Negative Indexes Support

bl4ckst0ne:
Hey,
I've added support for negative indexes. Let me know if there are any problems with that. 😉

## Feature Overview
Currently, negative integer indexes are not supported. It might be useful for obtaining one of the last elements when the list length is unknown. For example, when parsing a payload from the network / user input. In my case, it was a sorted list from the network that I wanted the second item from the end.

A basic usage might look like this: for `{'a': [1,2,3]}`, the path `a.[-1]` should return 3. 

## Implementation
To support negative indexes, `to_path` should parse `[negative index]` as a number. Thus, we need to alter just the regex.
The change is pretty small: just add `-?` before `\d+`. That way the minus sign is optional, so both positive and negative numbers will be matched.

## Tests
I've added cases with negative indexes for `to_path` function as well as `get` function.


dgilland (approved):
Good stuff!

dgilland:
Thanks for working on this!

PR #183, opened 2023-03-16
Better get typing for basic list usage

DeviousStoat:
`get` is very hard to type but a very nice simple usage of the function is to maybe get a value from a list without worrying about index errors, similarly to `get` method for `dict`.
Since using the function on a list with an `int` path is unequivocal, this usage can be typed. This is the goal of this PR.

### DIFF
--- a/src/pydash/objects.py
+++ b/src/pydash/objects.py
@@ -833,6 +833,21 @@ def for_in_right(obj, iteratee=None):
     return obj
 
 
+@t.overload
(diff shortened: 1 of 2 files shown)


dgilland (approved):

dgilland:
@DeviousStoat One thing I realized with the `get`is that it supports any object. If the passed in object is not a `dict` or `list`, then it falls back to using `getattr()` on the object.

Issue #184, opened 2023-03-18
Use sphinx-autodoc-typehints to autogenerate typed API docs and remove typing from docstrings


dgilland:
The new type annotations can replace the typing in the docstrings so that we don't have to maintain two sources of typing. All that's needed is to add `sphinx-autodoc-typehints` to the dev requirements in `setup.cfg` and then remove all typing from docstrings.

dgilland:
@DeviousStoat is this something you wanted to take on?

PR #186, opened 2023-03-18
Typing docs

DeviousStoat:
https://github.com/dgilland/pydash/issues/184

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,6 +58,7 @@ dev =
     tox
     twine
     wheel
+    sphinx-autodoc-typehints
 
 
 [bdist_wheel]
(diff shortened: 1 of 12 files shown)


dgilland (changes requested):
Beautiful! 💯

I did noticed that `helpers.py` has some docstring-types in `base_get` and `base_set`. I think we can just remove those since they aren't very helpful (and probably out of date anyway).

Everything else LGTM! 👍

DeviousStoat:
Oh yeah I thought I would leave them since these functions are not typed. But you are right, they are not very helpful. I removed them

dgilland (approved):

Issue #187, opened 2023-03-18
`to_dict` with list of length 2 strings


DeviousStoat:
I found a behaviour of `to_dict` that might not be exactly what we would expect.
The function first tries to use the builtin `dict` class for transformation and this class will try to build a dict if we pass a list of 2 elements iterators. This yields a funny result when we pass a list of length 2 strings:
```py
>>> import pydash
>>> pydash.to_dict(['hi', 'jo'])
{'h': 'i', 'j': 'o'}
```

Whereas I would expect this to return `{0: 'hi', 1: 'jo'}` as it would do if we pass a list of strings of length != 2:
```py
>>> pydash.to_dict(['hey', 'joe'])
{0: 'hey', 1: 'joe'}
```

Is this something that you would want to keep or can I try to make a PR to fix this?

dgilland:
It's been awhile since I looked at that function but if a >2 element list is returned with indexes as keys, then it should be the same for a 2 element list.

The next release with type annotations will be a major breaking change anyway so fine to fix this and not worry if this behavior was being relied on. Will just note it in the changelog.

dgilland:
Addressed in #188

PR #188, opened 2023-03-19
update to_dict types

DeviousStoat:
https://github.com/dgilland/pydash/issues/187

I looked into this but I think there isn't a good way to check this if we want to support iterable. We would have to iterate over the iterable which wouldn't be very nice. Since it is a very rare edge case (happens only if it is a iterable of strings and all of them have exact length of 2) and it is the behaviour of the builtin `dict` class, maybe it is not worth supporting.  What do you think?

This PR just adds the other types for `to_dict` which are supported by builtin `dict` and the iterable case implemented in this library.


src/pydash/objects.py
```diff
@@ -1920,7 +1920,17 @@ def to_dict(obj: t.Dict[T, T2]) -> t.Dict[T, T2]:
 
 
 @t.overload
-def to_dict(obj: t.List[t.Tuple[T, T2]]) -> t.Dict[T, T2]:
+def to_dict(obj: t.Iterable[t.Tuple[T, T2]]) -> t.Dict[T, T2]:
+    ...
+
+
+@t.overload
+def to_dict(obj: t.Iterable[t.List[str]]) -> t.Dict[str, str]:  # type: ignore
```
dgilland, line 1928: Curious about the `type: ignore` here. Is the typing not valid or what's going on that requires it? And I guess even with that ignore, the typing still gets applied?

DeviousStoat: mypy errors on overloaded signatures that overlap and have incompatible return types like the iterable ones do in this case because the type variable can match anything including the string list and tuples and the types of the dict returns are fixed and incompatible. More info [here](https://mypy.readthedocs.io/en/stable/more_types.html#type-checking-the-variants).

This could yield wrong types if the user's types are not narrow enough. For example we can port the example in the mypy docs to this case:
```py
import pydash

x: t.List[object] = [('a', 'b'), ('c', 'd')]  # this is right because tuple is an object too

reveal_type(to_dict(x))  # Revealed type is "builtins.dict[builtins.int, builtins.object]"
```

This is wrong, it matches the last overload because they overlap.

Unfortunately there is no way to make a type excluding other types in the python type system currently so I don't think there is a more correct way to type this. We could remove the `Iterable[T]` case to be more on the safe side so that it falls back to the last case and we still get a `Dict` type without the key and value types. 


> And I guess even with that ignore, the typing still gets applied?

And yes otherwise, the types still get applied, this just ignores the check of the overload itself, but when using it, it will still be applied as is validated in the type tests I added

dgilland: I'm not sure this typing, `def to_dict(obj: t.Iterable[t.List[str]]) -> t.Dict[str, str]:`, is valid, though, since it outputs `t.Dict[str, str]` conditionally based on the number of elements in the given list:

```python
>>> pydash.to_dict([['a'], ['c']])
{0: ['a'], 1: ['c']}
>>> pydash.to_dict([['a', 'c']])
{'a': 'c'}
>>> pydash.to_dict([['a', 'c'], ['x', 'y']])
{'a': 'c', 'x': 'y'}
>>> pydash.to_dict([['a', 'c', 'd'], ['x', 'y', 'z']])
{0: ['a', 'c', 'd'], 1: ['x', 'y', 'z']}
```

DeviousStoat: Yeah it is tricky to type. `dict` and `dict(iterator)` return very different results based on untypable conditions.

But my logic for this overload comes from the builtin types of `dict` [here](https://github.com/python/typeshed/blob/main/stdlib/builtins.pyi#L1035)
```py
    # Next overload is for dict(string.split(sep) for string in iterable)
    # Cannot be Iterable[Sequence[_T]] or otherwise dict(["foo", "bar", "baz"]) is not an error
    @overload
    def __init__(self: dict[str, str], __iterable: Iterable[list[str]]) -> None: ...
```

Which is indeed a very useful case to have typed. Here we can have something else instead of an error, so yeah I am not sure.

dgilland: I think the issue is that `pydash.to_dict` is sometimes `dict(obj)` and sometimes `dict(iterator(obj))` where the typing `def to_dict(obj: t.Iterable[t.List[str]]) -> t.Dict[str, str]:` is valid for `dict(obj)` but not for `dict(iterator(obj))`.

From the comment you posted:

> otherwise dict(["foo", "bar", "baz"]) is not an error

is true for `dict()` but not `dict(iterator(obj))` because of:

```python
def to_dict(obj):
    try:
        return dict(obj)
    except Exception:
        return dict(iterator(obj))
```

I think if we change that to:

```python
def to_dict(obj):
    return dict(iterator(obj))
```

then it would fix #187 and possibly make the typing for `pydash.to_dict` more consistent (maybe?).

With that change, we'd have:

```python
>>> pydash.to_dict(['hi', 'jo'])
{0: 'hi', 1: 'jo'}
>>> pydash.to_dict([['a', 'c', 'd'], ['x', 'y', 'z']])
{0: ['a', 'c', 'd'], 1: ['x', 'y', 'z']}
>>> pydash.to_dict([['a', 'c']])
{0: ['a', 'c']}
>>> pydash.to_dict([['a', 'c'], ['x', 'y']])
{0: ['a', 'c'], 1: ['x', 'y']}
```

DeviousStoat: Yes exactly. Using only `dict(iterator(obj))` would make it more consistent and a lot easier to type but it wouldn't work for the `dict` cases like list of 2 length tuple or list.

I believe it was the previous behaviour until you change it here:
```
    .. versionchanged:: 4.7.1
        Try to convert to ``dict`` using ``dict()`` first, then fallback to using
        ``pydash.helpers.iterator``.
```

dgilland: I did do that, didn't I.

Does seem kind of strange, though, that `pydash.to_dict([['a', 'b'], ['c', 'd']]) -> {'a': 'b', 'c': 'd'}` produces a different format than `pydash.to_dict([['a', 'b'], ['c', 'd'], ['e']]) -> {0: ['a', 'b'], 1: ['c', 'd'], 2: ['e']}`. Not sure if I realized that at the time.

So I guess part of the issue is that `def to_dict(obj: t.Iterable[t.List[str]]) -> t.Dict[str, str]:` is fine for `dict()` since it throws an error when the list of strings has a count != 2 so that when it's not an error, the `t.Dict[str, str]` type is valid. But for pydash, it happily falls back to returning `t.Dict[int, t.List[str]]`.

I guess my issue with having that particular type is that it's only sometimes right. Almost seems like ignoring that case would be better (i.e. have the types be less exact in favor of potentially being wrong). Or does the overall typing here able to distinguish between `pydash.to_dict([['a', 'b'], ['c', 'd']]) -> {'a': 'b', 'c': 'd'}` and `pydash.to_dict([['a', 'b'], ['c', 'd'], ['e']]) -> {0: ['a', 'b'], 1: ['c', 'd'], 2: ['e']}`?

The more I think about this function, the less useful I think it is given these inconsistencies. Like it's not very reliable for certain inputs and gives different dictionary keys based on some not so obvious properties of the input. Perhaps it should be removed in the next release.

DeviousStoat: Yes I completely agree that it is not very nice typing. But I don't think there is a way to distinguish between the two cases you shared.

It is relied on by `omit_by` and `pick_by` and the other functions that rely on these. And I guess these will have to use `dict(iterator(obj))` instead if we want it to work on all objects. So maybe just going back to the old implementation of `to_dict` is the best solution.

Or we could simply remove this overload indeed, it is not too bad

dgilland: After thinking about this, I'd like to remove the `dict()` call and just have `return dict(iterator(obj))`.

DeviousStoat: I changed the behaviour accordingly and rebased and fix types

dgilland (approved):

Issue #189, opened 2023-03-20
pytest_mypy_testing tests are very slow


dgilland:
Tests in `tests/pytest_mypy_testing` run very slowly:

```
$ pytest tests
...
2531 passed, 1 skipped in 109.87s (0:01:49)
```

compared to running without those tests:

```
$ pytest tests -k "not mypy_testing"
...
1835 passed, 1 skipped, 696 deselected in 2.08s
```

It's even worse when running these tests with `inv test` which ended up being `262.77s`.

@DeviousStoat have you experienced similar timings? Do you know why it's happening? I'm assuming it's something to do with how pytest-mypy-testing is checking the types. This increase in test time is not desirable. If there isn't an easy way to speed it up, then we may need to remove the pytest-mypy-testing tests or move their execution out of the main test suite (possibly just run on demand or as part of some other process that doesn't run during CI).

DeviousStoat:
Yes unfortunately the library is very slow but this is the only one that does this that I found. I planned to look into it and try to fix it on their side but I haven't yet. Yes it would be easy to skip them so it would probably be better to keep them around

dgilland:
Mitigated in #191 by making tests optional.

PR #191, opened 2023-04-01
Make mypy tests optional

DeviousStoat:
https://github.com/dgilland/pydash/issues/189

mypy tests are very slow. This makes them optional so we can still run them locally with `inv test --with-mypy-tests` but by default they won't be ran, this includes in the CI


dgilland (approved):
Awesome!

PR #194, opened 2023-04-22
Export all from `pydash`

DeviousStoat:
Using pyright as a type checker with `reportPrivateUsage=true` we get errors that the objects are not exported from `pydash`.
The reason is explained [here](https://github.com/microsoft/pylance-release/issues/2953#issuecomment-1168408943) by pyright's maintainer.


dgilland:
Is it possible to have a test that would detect a missing `__all__` item?

DeviousStoat:
Yes of course. We can reuse the logic of one function from the chain interface generator script that parses the imports.
But I believe some linters can detect unused imports as well. This would detect missing `__all__` item as well. Maybe in a more elegant manner.
I will try to look into it later today.

DeviousStoat:
The `F401` rule from `flake8` that handles this was disabled. I enabled it back. It is nice, it understands when we export it in `__all__` too so it adds the test you requested.
For example if I remove `"chunk"` from the `__all__` list in `src/pydash/__init__.py` I get a lint error from `flake8`:
```sh
src/pydash/__init__.py:5: [F401] '.arrays.chunk' imported but unused
```

dgilland:
Thanks for doing this!

DeviousStoat:
Thank you! Could we get a release with this fix? This is a bit blocking for some of my projects

dgilland:
Released v7.0.2: https://pypi.org/project/pydash/7.0.2/

PR #196, opened 2023-05-02
Type fixes

DeviousStoat:
The iteratee of all these functions should be able to return `Any` type so that the criterion can be any subvalue of the elements.

The other `_by` functions seem fine. It is the `arrays` one that had this problem.


dgilland:
Released this and https://github.com/dgilland/pydash/pull/195 in v7.0.3: https://pypi.org/project/pydash/7.0.3/

Issue #197, opened 2023-05-23
Incompatible with typing-extensions 4.6


lephuongbg:
Got this error when upgrading typing-extensions to 4.6. pydash works fine with typing-extensions 4.5.

```
utils/__init__.py:20: in <module>
    from pydash import last
.venv/lib/python3.10/site-packages/pydash/__init__.py:170: in <module>
    from .objects import (
.venv/lib/python3.10/site-packages/pydash/objects.py:17: in <module>
    from .utilities import PathToken, to_path, to_path_tokens
.venv/lib/python3.10/site-packages/pydash/utilities.py:577: in <module>
    class MemoizedFunc(Protocol[P, T, T2]):
../lang/lib/python3.10/typing.py:312: in inner
    return func(*args, **kwds)
.venv/lib/python3.10/site-packages/typing_extensions.py:672: in __class_getitem__
    raise TypeError(
E   TypeError: Parameters to Protocol[...] must all be type variables. Parameter 1 is ~P
```

Tenzer:
I did some bisecting of the changes between typing-extensions 4.5.0 and 4.6.0 and found the changes in https://github.com/python/typing_extensions/pull/137 to be the culprit.

I have posted on that PR for any pointers on how to fix this incompatibility, as I can't immediately work it out.

Tenzer:
typing-extensions 4.6.1 has been released now which fixes this.

jenstroeger:
With the typing-extensions package’s fix, shouldn’t pydash also change its own requirements to _avoid_ importing the culprit version — [this requirement](https://github.com/dgilland/pydash/blob/051fe69c3e523f903a0a0ea6ca6a5c2d4b83a3e7/setup.cfg#L33C2-L34) 
```python
install_requires =
    typing-extensions>=3.10
```
seems rather loose to me.

PR #198, opened 2023-05-30
fix(deps): exclude buggy typing-extensions 4.6

jenstroeger:
That’s to avoid that a user accidentally (through other deps) installs the incompatible/buggy typing-extensions package [v4.6.0](https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#release-460-may-22-2023), see also issue https://github.com/dgilland/pydash/issues/197.

This should be a bug-fix release for pydash (assuming that pydash follows [SemVer](https://semver.org/)) so that other packages which depend on pydash can respond accordingly.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ packages = find:
 python_requires = >=3.7
 include_package_data = True
 install_requires =
-    typing-extensions>=3.10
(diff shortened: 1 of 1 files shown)


dgilland:
Thanks for the fix!

PR #207, opened 2023-09-26
Correct function overloading for map_keys

Drvanon:
There is a small bug in `Chain.map_keys`. When the iteratee accepts only one argument, the type of the argument is said to be of type `T2`, which is the type of the values. However since `map_keys` well, maps keys, the type here should be `T`. This commit should fix that.


dgilland:
This is the intended behavior: https://github.com/dgilland/pydash/issues/208#issuecomment-1735724171

PR #214, opened 2024-01-19
Fix path behaviour for set_ 

### LINKED ISSUE #213: Unexpected difference in path behavior for set_() when accessing attribute of object
Hello,

Thank you for a great library!

I came across a discrepancy when using the set_ function where using a string path and a list path behaves differently:

```
❯ python
Python 3.11.6 (main, Oct  2 2023, 13:45:54) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pydash
>>> pydash.__version__
'7.0.6'
>>> class A:
...     def __init__(self):
...             self.x = {}
...
>>> a = A()
>>> pydash.set_(a, 'x.a.b', 1)
<__main__.A object at 0x7f3e010a17d0>
>>> a.x
{'a': {'b': 1}}
>>> pydash.set_(a, ['x', 'y', 'z'], 1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/venv/lib/python3.11/site-packages/pydash/objects.py", line 1825, in set_
    return set_with(obj, path, value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/pydash/objects.py", line 1856, in set_with
    return update_with(obj, path, pyd.constant(value), customizer=customizer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/pydash/objects.py", line 2453, in update_with
    base_set(target, last_key, callit(updater, value))
  File "/venv/lib/python3.11/site-packages/pydash/helpers.py", line 219, in base_set
    key = int(key)
          ^^^^^^^^
ValueError: invalid literal for int() with base 10: 'z'
>>> a.x
{'a': {'b': 1}, 'y': []}
```

The expected behavior would be for `a.x` to equal `{'a': {'b': 1}, 'y': {'z': 1 }}` in the end.

Note that this issue does not occur when accessing e.g. a dict directly:

```
>>> a = {}
>>> pydash.set_(a, 'x.a.b', 1)
{'x': {'a': {'b': 1}}}
>>> pydash.set_(a, ['x', 'y', 'z'], 1)
{'x': {'a': {'b': 1}, 'y': {'z': 1}}}
```

DeviousStoat:
fixes #213

The default type when setting a missing key is usually given by the `PathToken` built in `to_path_tokens`.
However this function is incomplete and only builds a list of `PathToken`s if the path is a string.

The issue here is that that [line](https://github.com/dgilland/pydash/blob/develop/src/pydash/objects.py#L2408) sets the default type to `list` if the object to update is not a `dict`.
And as stated above, when we have a path as a list `to_path_tokens` just returns the list and we don't have any information about the default factory, we just default to `list`.
So when we create the first list and try to update it we expect an index.

This PR updates the `to_path_tokens` function to be more general and build the list of `PathToken` even for a list.

The added `pragma: no cover` are there because this case should now be impossible to reach, the only case where the tokens are not `PathToken`s is when the path is only 1 non string value and we go directly to the end case as the for loop only iterates on the initial values.

### DIFF
--- a/src/pydash/objects.py
+++ b/src/pydash/objects.py
@@ -2295,9 +2295,6 @@ def update_with(obj, path, updater, customizer=None):  # noqa: C901
     default_type = dict if isinstance(obj, dict) else list
     tokens = to_path_tokens(path)
 
-    if not pyd.is_list(tokens):  # pragma: no cover
-        tokens = [tokens]
-
     last_key = pyd.last(tokens)
 
     if isinstance(last_key, PathToken):
@@ -2367,9 +2364,6 @@ def unset(obj: t.Union[t.List, t.Dict], path: PathT) -> bool:  # noqa: C901
     """
     tokens = to_path_tokens(path)
 
-    if not pyd.is_list(tokens):  # pragma: no cover
-        tokens = [tokens]
-
     last_key = pyd.last(tokens)
 
     if isinstance(last_key, PathToken):
@@ -2397,10 +2391,12 @@ def unset(obj: t.Union[t.List, t.Dict], path: PathT) -> bool:  # noqa: C901
     if target is not UNSET:
         try:
             try:
-                target.pop(last_key)
+                # last_key can be a lot of things
+                # safe as everything wrapped in try/except
+                target.pop(last_key)  # type: ignore
                 did_unset = True
             except TypeError:
-                target.pop(int(last_key))
+                target.pop(int(last_key))  # type: ignore
                 did_unset = True
         except Exception:
             pass
--- a/src/pydash/objects.py
+++ b/src/pydash/objects.py
@@ -2309,7 +2309,7 @@ def update_with(obj, path, updater, customizer=None):  # noqa: C901
         if isinstance(token, PathToken):
             key = token.key
             default_factory = pyd.get(tokens, [idx + 1, "default_factory"], default=default_type)
-        else:
+        else:  # pragma: no cover
             key = token
             default_factory = default_type
 
@@ -2384,7 +2384,7 @@ def unset(obj: t.Union[t.List, t.Dict], path: PathT) -> bool:  # noqa: C901
     for token in pyd.initial(tokens):
         if isinstance(token, PathToken):
             key = token.key
-        else:
+        else:  # pragma: no cover
             key = token
 
         try:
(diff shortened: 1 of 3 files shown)


src/pydash/objects.py
```diff
@@ -2422,7 +2422,7 @@ def update_with(obj, path, updater, customizer=None):  # noqa: C901
         if isinstance(token, PathToken):
             key = token.key
             default_factory = pyd.get(tokens, [idx + 1, "default_factory"], default=default_type)
-        else:
+        else:  # pragma: no cover
```
dgilland, line 2425: If the `else` case is no longer valid and `isinstance(token, PathToken)` is always true, then I think we could remove the if-statement and just have whatever is under the main `if` section.

Unless there was a reason to leave them in?

DeviousStoat: Yeah makes sense. I removed them.

I left them just to be safe, but I guess if it happens it would probably be a rare edge case. The test cases cover a lot of cases and every passes so it should be safe enough.

DeviousStoat:
Lint fails in CI on files I didn't touch and everything passes localy. Not sure what's going on. Maybe we can just try rerunning it?

dgilland:
> Lint fails in CI on files I didn't touch and everything passes localy. Not sure what's going on. Maybe we can just try rerunning it?

Latest version of black made some changes to the formatting and required some flake8 ignores on certain rules. Rebase with `develop` to get those changes and lint should be good.

src/pydash/objects.py
dgilland: > would probably be a rare edge case

From looking at the code, the `tokens` are from `to_path_tokens()` which always returns a `PathToken` list, right? So should never be any other type. Unless I missed something in `to_path_tokens()` where a different type could be included?

DeviousStoat: `to_path_tokens` can return something else than a `PathToken` if it is a single element like `"0"` of `"some_key"`.
But I guess we should probably make `to_path_tokens` always return a list of `PathToken` all the time. It would be more consistant and easier to reason about.
I added a commit with it. Let me know what you think

dgilland (approved):
Nice work!

PR #216, opened 2024-02-02
Add curried version of the multiple argument predicate functions

DeviousStoat:
This adds curried version of the predicate functions accepting multiple arguments.

So you can nicely do stuff like:
```py
import pydash as pyd

assert pyd.filter_([1, 2, 3], pyd.gt_pred(2)) == [3]
assert pyd.filter_([1, 2, 3], pyd.in_range_pred(2, 3)) == [2]
```

Instead of having to write the whole function or manually currying:
```py
import pydash as pyd

assert pyd.filter_([1, 2, 3], lambda value: value > 2) == [3]
assert pyd.filter_([1, 2, 3], pyd.curry_right(pyd.in_range)(3)(2)) == [2]
```

I am not very satisfied with the naming though, let me know what you think.


dgilland:
> I am not very satisfied with the naming though, let me know what you think.

Instead of suffixing with `_pred`, I'd suggest using `_cmp` (short for comparator).

DeviousStoat:
> > I am not very satisfied with the naming though, let me know what you think.
> 
> Instead of suffixing with `_pred`, I'd suggest using `_cmp` (short for comparator).

Makes more sense, thank you. I changed it.

dgilland (approved):

PR #218, opened 2024-02-09
Fix call_math_operator on false values

### LINKED ISSUE #217: multiply does not math
```
from pydash import py_

print(py_.multiply(0, 1) == 0)
```

returns false instead of true (because its returning 1 instead of 0)

DeviousStoat:
Fixes #217 

`0` is a `False` value hence checking for `not value` will return `True` and we will fallback to the default value.

### DIFF
--- a/tests/test_numerical.py
+++ b/tests/test_numerical.py
@@ -200,7 +200,7 @@ def test_moving_mean(case, expected):
 
 @parametrize(
     "multiplier,multiplicand,expected",
-    [(10, 5, 50), (None, 1, 1), (None, None, 1), (1.5, 3, 4.5), (-10, 2, -20)],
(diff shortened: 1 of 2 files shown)


tests/test_numerical.py
```diff
@@ -200,7 +200,7 @@ def test_moving_mean(case, expected):
 
 @parametrize(
     "multiplier,multiplicand,expected",
-    [(10, 5, 50), (None, 1, 1), (None, None, 1), (1.5, 3, 4.5), (-10, 2, -20)],
+    [(10, 5, 50), (None, 1, 1), (None, None, 1), (1.5, 3, 4.5), (-10, 2, -20), (0, 1, 0)],
```
slallum, line 203: I suggest adding the case of `(1, 0, 0)`

slallum (approved):

dgilland (approved):

PR #232, opened 2024-07-21
Type fixes

### LINKED ISSUE #231: mypy overload-overlap errors starting with mypy==1.11.0
Starting with mypy==1.11.0, mypy throws the following errors:

```
src/pydash/objects.py:924:1: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [overload-overlap]
src/pydash/objects.py:962:1: error: Overloaded function signatures 1 and 3 overlap with incompatible return types  [overload-overlap]
src/pydash/objects.py:966:1: error: Overloaded function signatures 2 and 4 overlap with incompatible return types  [overload-overlap]
src/pydash/objects.py:1509:1: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [overload-overlap]
src/pydash/objects.py:1555:1: error: Overloaded function signatures 1 and 4 overlap with incompatible return types  [overload-overlap]
src/pydash/objects.py:1559:1: error: Overloaded function signatures 2 and 5 overlap with incompatible return types  [overload-overlap]
src/pydash/objects.py:1948:1: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [overload-overlap]
src/pydash/objects.py:2139:1: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [overload-overlap]
```

Affected functions are:

- `invert`
- `invert_by`
- `omit`
- `omit_by`
- `to_dict`
- `to_pairs`

@DeviousStoat any ideas for how to fix these?

DeviousStoat:
fixes #231


DeviousStoat:
I have no idea why we get these mypy syntax errors though. Doesn't even seem to point to the right line and I can't seem to be able to reproduce locally. I will try to figure it out later today

dgilland:
I can reproduce it locally if I delete the `.mypy_cache` folder and then run mypy (but only on the first run):

```
$ rm -r .mypy_cache
$ inv mypy
src/pydash/numerical.py:104:68: error: X | Y syntax for unions requires Python 3.10  [syntax]
src/pydash/arrays.py:104:68: error: X | Y syntax for unions requires Python 3.10  [syntax]
src/pydash/objects.py:104:68: error: X | Y syntax for unions requires Python 3.10  [syntax]
$ inv mypy
$
```

I don't think there's even any `X | Y` type syntax being used either, which makes me think it might be an issue with mypy.

DeviousStoat:
I think I found the issue. It happens when using the `SupportsRichComparisonT` type variable.

It is defined in the python base library typesheds. When we are using a >=3.10 python version mypy will use the >=3.10 typesheds and the definition is
```py
SupportsRichComparison: TypeAlias = SupportsDunderLT[Any] | SupportsDunderGT[Any]
SupportsRichComparisonT = TypeVar("SupportsRichComparisonT", bound=SupportsRichComparison)
```
And in the `pyproject.toml` `mypy` section we have `python_version = "3.8"` so it fails when type checking using the 3.10 typeshed with the `X | Y` union syntax that is used in the typeshed. If we set `python_version = "3.10"` it works. But I don't think we want that, we probably want mypy to check using the lowest supported python version.

One other thing that fixes it is adding `from __future__ import annotations` on top of the affected modules. This allows using the `X | Y` syntax in <3.10 python version. It also postpones the evaluation of annotations so we could write `collection: t.Mapping[t.Any, SupportsRichComparisonT], default: Unset = UNSET` instead of `collection: t.Mapping[t.Any, "SupportsRichComparisonT"], default: Unset = UNSET` for type variable that are imported only during `TYPE_CHECKING`.

Otherwise we could redefine the `SupportsRichComparison` values inside the library.

I think the `from __future__ import annotations` is probably the best solution, it's a good practice to support different python versions with mypy. Let me know what you think.

dgilland:
From future import seems best.

DeviousStoat:
neat, everything passes now :D

dgilland (approved):

PR #241, opened 2026-05-25
fix: support iterables in take_while and drop_while

### LINKED ISSUE #204: Take while does not support generators
If I add a generator as the object of my chain and I use the `take_while` function, it tries to call the nth index of the generator which is not supported. 

A suggesteion for this would be 
```python
def take_while(array, predicate=None):
    array = list(array)
    n = 0
    for is_true, _, _, _ in iteriteratee(array, predicate):
         if is_true:
             n += 1
         else:
             break
      return array[:n]
```
or
```python
def take_while(array, predicate=None):
    new_array = []
    while True:
         is_true, n, _, _ in iteriteratee(array, predicate):
         if not is_true:
             break  
         new_array.append(array[n])
    return new_array
```

SAY-5:
`take_while` and `drop_while` (and their callers `take`/`drop`) raise `TypeError: 'generator' object is not subscriptable` when passed a generator or other non-subscriptable iterable, because both functions slice the original `array` after counting. Converting to a list first fixes the crash on valid input.

Fixes #204.

### DIFF
--- a/src/pydash/arrays.py
+++ b/src/pydash/arrays.py
@@ -449,14 +449,15 @@ def drop_while(array, predicate=None):
 
     .. versionadded:: 1.1.0
     """
+    arr = list(array)
     n = 0
-    for is_true, _, _, _ in iteriteratee(array, predicate):
+    for is_true, _, _, _ in iteriteratee(arr, predicate):
         if is_true:
             n += 1
         else:
             break
 
-    return array[n:]
+    return arr[n:]
 
 
 def duplicates(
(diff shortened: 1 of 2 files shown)


src/pydash/arrays.py
```diff
@@ -449,14 +449,15 @@ def drop_while(array, predicate=None):
 
     .. versionadded:: 1.1.0
     """
+    arr = list(array)
     n = 0
-    for is_true, _, _, _ in iteriteratee(array, predicate):
+    for is_true, _, _, _ in iteriteratee(arr, predicate):
```
dgilland: This should probably build a new list from true elements instead of creating a full copy of the original.

src/pydash/arrays.py
```diff
@@ -2093,14 +2094,15 @@ def take_while(array, predicate=None):
...
```
dgilland: Similar to above, this should probably build a new list with the remaining elements once the predicate becomes false.

dgilland (changes requested):

PR #247, opened 2026-06-24
Preserve empty keys in deep paths

CodingFeng101:
## Summary

This preserves empty string keys when parsing deep path strings that contain consecutive dot delimiters, such as ..b.

Before this change, 	o_path_tokens() used ilter(None, ...) on the regex split output. That removed both structural empty strings around list-index tokens and meaningful empty string path keys. As a result, ..b was parsed as ['a', 'b'], so helpers like get() and set_() skipped an empty-string dictionary key.

## Validation

- Reproduced before the fix:
  - 	o_path('a..b') returned ['a', 'b']
  - get({'a': {'': {'b': 1}}}, 'a..b', 'missing') returned 'missing'
  - set_({}, 'a..b', 1) produced {'a': {'b': 1}}
- After the fix:
  - 	o_path('a..b') returns ['a', '', 'b']
  - get({'a': {'': {'b': 1}}}, 'a..b') returns 1
  - set_({}, 'a..b', 1) produces {'a': {'': {'b': 1}}}
- python -m pytest -o addopts='' tests/test_utilities.py::test_to_path tests/test_objects.py::test_get tests/test_objects.py::test_set_ -q -> 65 passed
- python -m pytest -o addopts='' tests --ignore=tests/pytest_mypy_testing -q -> 1872 passed
- python -m ruff check src tests --exclude tests/pytest_mypy_testing -> passed
- git diff --check -> passed

I also tried python -m pytest -o addopts='' -q, but that treats 	ests/pytest_mypy_testing as regular Python and fails on expected 
eveal_type NameErrors. The regular test suite passes when that mypy-specific directory is ignored.

### DIFF
--- a/src/pydash/utilities.py
+++ b/src/pydash/utilities.py
@@ -1441,13 +1441,38 @@ def _to_path_token(key) -> PathToken:
     )
 
 
+def _to_path_keys(value):
+    keys = []
+    parts = RE_PATH_KEY_DELIM.split(value)
+
+    for idx, key in enumerate(parts):
+        if key is None:
+            continue
+
+        if key == "":
+            prev_part = pyd.get(parts, idx - 1) if idx else None
+            next_part = pyd.get(parts, idx + 1)
+            prev_is_list_index = _maybe_list_index(prev_part) is not None
+            next_is_list_index = _maybe_list_index(next_part) is not None
+
+            if prev_is_list_index or next_is_list_index:
+                continue
+
+            if prev_part is not None and next_part is not None:
+                continue
+
+        keys.append(key)
+
+    return keys
+
+
 def to_path_tokens(value) -> t.List[PathToken]:
     """Parse `value` into :class:`PathToken` objects."""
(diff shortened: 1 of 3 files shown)


src/pydash/utilities.py
```diff
@@ -1441,13 +1441,38 @@ def _to_path_token(key) -> PathToken:
...
+            prev_part = pyd.get(parts, idx - 1) if idx else None
+            next_part = pyd.get(parts, idx + 1)
+            prev_is_list_index = _maybe_list_index(prev_part) is not None
+            next_is_list_index = _maybe_list_index(next_part) is not None
+
+            if prev_is_list_index or next_is_list_index:
+                continue
+
+            if prev_part is not None and next_part is not None:
+                continue
```
dgilland, line 1462: This doesn't distinguish between existing values that are `None` and a missing value that was default to `None`. Internally, functions use the `UNSET` sentinel to disambiguate (i.e. `get(..., default=UNSET)`).

dgilland (changes requested):

PR #250, opened 2026-06-29
Fix to_list() returning dict_values instead of a list for dicts (#249)

### LINKED ISSUE #249: to_list() on a dict returns dict_values, not a list
### Description

`pydash.to_list()` on a dict returns a `dict_values` view object, not an actual `list`. A function named `to_list` returning a non-list breaks `isinstance(result, list)` and any code that indexes/concatenates the result.

### Reproduction

```python
import pydash

r = pydash.to_list({"a": 1, "b": 2})
print(repr(r))                 # dict_values([1, 2])
print(isinstance(r, list))     # False
```

### Expected vs actual

- expected: `[1, 2]` (a real `list`)
- actual: `dict_values([1, 2])`

### Root cause

For a dict input `to_list` yields `obj.values()` without materializing it into a list. Wrapping the result in `list(...)` restores the documented contract ("Creates a list from another object").

HarperZ9:
Fixes #249.

## Problem

`to_list()` returns a `dict_values` view rather than a `list` when given a dict:

```python
>>> import pydash as _
>>> _.to_list({"a": 1, "b": 2})
dict_values([1, 2])          # not a list
>>> isinstance(_.to_list({"a": 1, "b": 2}), list)
False
```

Every other branch of `to_list` returns a real `list`, the docstring says "Converts ... to a list", and the typed signature (and the mypy `reveal_type` tests for `to_list`) promise `list[T]`. Only the dict branch (`return obj.values()`) breaks the contract.

## Fix

`return list(obj.values())`.

## Test

Added `test_to_list` in `tests/test_objects.py` covering dict, list, tuple, and scalar inputs, asserting both the value and `isinstance(result, list)`. Verified the dict case fails on the current code (`assert dict_values([1, 2, 3]) == [1, 2, 3]`) and passes with the fix.

Local verification on 2026-07-02 with Python 3.12.10:

```console
$env:PYTHONPATH='src'; python -m pytest tests/test_objects.py --no-cov
290 passed
```

Happy to add a CHANGELOG entry if you'd like; I left it out since the changelog looks maintainer-managed per release.

### DIFF
--- a/src/pydash/objects.py
+++ b/src/pydash/objects.py
@@ -2115,7 +2115,7 @@ def to_list(obj, split_strings=True):
     if isinstance(obj, list):
         return obj[:]
     elif isinstance(obj, dict):
-        return obj.values()
+        return list(obj.values())
     elif not split_strings and isinstance(obj, (str, bytes)):
         return [obj]
     elif split_strings and isinstance(obj, bytes):
--- a/tests/test_objects.py
+++ b/tests/test_objects.py
@@ -181,6 +181,21 @@ def test_to_dict(case, expected):
 @parametrize(
     "case,expected",
     [
+        ({"a": 1, "b": 2, "c": 3}, [1, 2, 3]),
+        ([1, 2, 3], [1, 2, 3]),
+        ((1, 2), [1, 2]),
+        (1, [1]),
+    ],
+)
+def test_to_list(case, expected):
+    result = _.to_list(case)
+    assert result == expected
+    assert isinstance(result, list)
+
+
+@parametrize(
+    "case,expected",
+    [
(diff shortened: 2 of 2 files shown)


HarperZ9:
Fresh local verification for this no-CI PR, from a Windows checkout of head `7ecda7bd78195d4fa5658cfae816116464e6c3aa` with Python 3.12.10:

- `$env:PYTHONPATH='src'; python -m pytest tests/test_objects.py::test_to_list --no-cov` -> 4 passed
- `$env:PYTHONPATH='src'; python -m pytest tests/test_objects.py --no-cov` -> 290 passed

I set `PYTHONPATH=src` so the tests exercise this checkout rather than any installed `pydash`. The checkout is clean after the run.

dgilland:
Thanks for the contribution, much appreciated! 👍

